PR #180, opened 2018-09-10
Support common practice of omitting the double slash in file urls.

nilsboy:
This makes the goto definition of javacs[1] work.

[1] Java language support for Visual Studio Code using javac
(https://github.com/georgewfraser/vscode-javac)

See Also:

- The // after the file: is part of the general syntax of URLs. (The double
slash // should always appear in a file URL according to the specification,
but in practice many Web browsers allow it to be omitted).
(https://en.wikipedia.org/wiki/File_URI_scheme#Meaning_of_slash_character)

- redundant syntax such as two slashes before an empty authority
(https://docs.oracle.com/javase/7/docs/api/java/net/URI.html)

### DIFF
--- a/autoload/lsp/utils.vim
+++ b/autoload/lsp/utils.vim
@@ -70,7 +70,7 @@ if has('win32') || has('win64')
     endfunction
 else
     function! lsp#utils#uri_to_path(uri) abort
-        return s:decode_uri(a:uri[len('file://'):])
+        return s:decode_uri(substitute(a:uri, '\v^file:/*', '/', 'g'))
     endfunction
 endif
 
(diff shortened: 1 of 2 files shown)


autoload/lsp/utils.vim
```diff
@@ -70,7 +70,7 @@ if has('win32') || has('win64')
     endfunction
 else
     function! lsp#utils#uri_to_path(uri) abort
-        return s:decode_uri(a:uri[len('file://'):])
+        return s:decode_uri(substitute(a:uri, '\v^file:/*', '/', 'g'))
```
mattn: If the uri contains hostname, the result should NOT be expected. This expect /path/to/the/file but the result which have hostname is /hostname/path/to/file.

mattn:
timedout

PR #193, opened 2018-10-21
Fix apply text edits

augustomelo:
The correct way to deal with edge cases (column = 0 or column = line length) is to use selection=excluise (does not delete the last character) and virutaledit=onemore (enable select last linecharacter) instead of transforming the text edits.

### DIFF
--- a/autoload/lsp/ui/vim.vim
+++ b/autoload/lsp/ui/vim.vim
@@ -484,10 +484,18 @@ function! s:apply_text_edits(uri, text_edits) abort
 
         try
             let l:was_paste = &paste
+            let l:was_selection = &selection
+            let l:was_virtualedit = &virtualedit
+
             set paste
+            set selection=exclusive
+            set virtualedit=onemore
+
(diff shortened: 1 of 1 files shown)


oblitum:
Is this related with `:LspRename`? Because `:LspRename` for variables named with a single character is not working.

augustomelo:
Thank you for your input, could you provide the logs for that?

oblitum:
@augustomelo ah, my bad, it's not related to identifier length. It doesn't rename if the file is not yet saved, for example, if you do one rename, don't save, and rename the same thing again to another name, it doesn't get renamed. I'm using [ccls](https://github.com/MaskRay/ccls) for c++. I'm unsure whether log for this is still needed? Maybe it's not supported. The interface is misleading though because it looks like it supports it. When I call `:LspRename` on an unsaved buffer over a variable previously renamed, it does offers the actual variable name to be renamed, but then doesn't rename anything upon execution.

PR #211, opened 2018-12-13
FZF support

ddcien:
FZF support has been integrated into my commits. Now you can handle location, symbol, and code action selections with FZF if it is loaded.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -1,6 +1,8 @@
 let s:enabled = 0
 let s:already_setup = 0
+let s:code_lens_tmr = 0
 let s:servers = {} " { lsp_id, server_info, init_callbacks, init_result, buffers: { path: { changed_tick } }
+
 
 let s:notification_callbacks = [] " { name, callback }
 
@@ -65,22 +67,22 @@ endfunction
 
 function! s:server_status(server_name) abort
     if !has_key(s:servers, a:server_name)
-        return "unknown server"
+        return 'unknown server'
     endif
     let l:server = s:servers[a:server_name]
     if has_key(l:server, 'exited')
-        return "exited"
+        return 'exited'
     endif
     if has_key(l:server, 'init_callbacks')
-        return "starting"
+        return 'starting'
     endif
     if has_key(l:server, 'failed')
-        return "failed"
+        return 'failed'
     endif
     if has_key(l:server, 'init_result')
-        return "running"
+        return 'running'
     endif
-    return "not running"
+    return 'not running'
 endfunction
 
 " Returns the current status of all servers (if called with no arguments) or
@@ -135,7 +137,7 @@ function! s:register_events() abort
         autocmd BufReadPost * call s:on_text_document_did_open()
         autocmd BufWritePost * call s:on_text_document_did_save()
         autocmd BufWinLeave * call s:on_text_document_did_close()
-        autocmd InsertLeave * call s:on_text_document_did_change()
+        autocmd InsertLeave,TextChanged * call s:on_text_document_did_change()
         autocmd CursorMoved * call s:on_cursor_moved()
     augroup END
     call s:on_text_document_did_open()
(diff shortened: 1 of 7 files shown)


mkwork:
Sounds quite cool!

autoload/lsp/ui/vim.vim
```diff
@@ -292,6 +484,21 @@ function! lsp#ui#vim#code_action() abort
...
+endfunction
+
+function! s:show_locations() abort
+    if !get(g:, 'lsp_fzf_enable') || !get(g:, 'loaded_fzf')
+        botright copen
+    else
+        let l:qfl = getqflist()
+        call map(l:qfl, {idx, item -> string(idx + 1) . '. ' . fnamemodify(expand(bufname(item['bufnr'])), ":~:.") . ':' . string(item['lnum']) . ':' . string(item['col']) . ':' . item['text']})
+        call fzf#run(fzf#wrap({'source': l:qfl, 'sink': function('s:jump_to_location'), 'options': '--reverse +m --prompt="Jump> "'}))
+    endif
```
mattn: Please reverse the condition. We may add yet-another selector in future.
```
if extensibie1
  call do_extend1()
  return
endif
if extensibie2
  call do_extend2()
  return
endif
call do_original()
```

ddcien: Yes, that makes sense. Fixed.

autoload/lsp.vim
```diff
@@ -135,7 +137,7 @@ function! s:register_events() abort
         autocmd BufReadPost * call s:on_text_document_did_open()
         autocmd BufWritePost * call s:on_text_document_did_save()
         autocmd BufWinLeave * call s:on_text_document_did_close()
-        autocmd InsertLeave * call s:on_text_document_did_change()
+        autocmd InsertLeave,TextChanged * call s:on_text_document_did_change()
```
prabirshrestha: would need to debounce `TextChanged` as it could slow down vim due to did change. When we implement incremental change it would be easier. But there is a PR currently for it at https://github.com/prabirshrestha/vim-lsp/pull/188

autoload/lsp/ui/vim.vim
```diff
@@ -377,30 +643,89 @@ function! s:handle_text_edit(server, last_req_id, type, data) abort
...
+        endif
+    endif
+
+    if l:ret == 0
         return
     endif
 
-    while l:index < len(l:codeActions)
-        call add(l:choices, string(l:index + 1) . ' - ' . l:codeActions[index]['title'])
+    if a:action['command'] ==# 'cquery._applyFixIt'
```
prabirshrestha: :( we shouldn't add lang server specific things here. Why is this needed?

autoload/lsp/ui/vim/utils.vim
```diff
@@ -54,6 +55,14 @@ let s:symbol_kinds = {
     \ '16': 'number',
     \ '17': 'boolean',
     \ '18': 'array',
+    \ '19': 'object',
```
prabirshrestha: this can be in a different PR.

prabirshrestha:
I'm getting closer to implementing a new [fuzzy search plugin in vim](https://github.com/prabirshrestha/quickpick.vim) with `on_change` support. The api will look something like this. (I haven't pushed the code yet)

```vim
function! s:on_change(id, value) abort
    " todo: properly debounce and cancel previous on-going search
    call s:search_npm(a:value, {items->quickpick#set_items(a:id, items)})
endfunction

function! s:on_accept(id, items) abort
    " use jobs for async install
    call quickpick#close(a:id)
    call job_start([&shell, &shellcmdflag, 'npm install ' . a:items[0]])
endfunction

let id = quickpick#create({
    \ 'on_change': function('s:on_change'),
    \ 'on_accept': function('s:on_accept'),
    \ 'auto_open': 0
    \ })
call quickpick#show(id)
```
Simple pickers one can directly using this api.

```vim
call quickpick#create({
    \ 'on_accept': function('s:on_accept'),
    \ 'items': ['item1', 'item2'],
    \ })
```

Then we can expose a extension point in vim-lsp that will look something like this.

```vim
call lsp#register_quickpick_ui('quickpick.vim', function('s:quickpick_vim'))
call lsp#register_quickpick_ui('ctrlp', function('s:ctrlp_quickpick'))
call lsp#register_quickpick_ui('fzf', function('s:fzf_quickpick'))
let g:lsp_default_quickpick_ui = 'fzf'
```

We need to figure out what the interface for `register_quickpick_ui` would like.

prabirshrestha:
Finally got a proof of concept for quickpick. Here is how it looks. This makes it lot more powerful than FZF/Ctrlp and would work well for document and specially workspace symbols where a project could have possible thousands of symbols and you would want the langserver to search instead of vim-lsp. quickpick.vim just acts as a dump ui. 

NPM Picker

All the above is implemented in pure vimscript and follows the footstep of vim-lsp and asyncomplete.vim

Note: quickpick.vim is very much work in progress https://github.com/prabirshrestha/quickpick.vim

mattn:
Hi @ddcien 

Thanks your contribution. Will you work about this PR ? If you don't, I'll continue to implement codeAction with your commits.

edganiukov:
I would love to have FZF support.

mattn:
IMO, I do not like to lock-in to 1 selector command.

prabirshrestha:
No plans to support FZF as first class support. There are lot of fuzzy search UI and I would like to avoid supporting N number of UIs. I personally use ctrlp and [vim-fz](https://github.com/mattn/vim-fz) with fzf executable which I find it lot better than `fzf.vim`.

```vim
if executable('fzf')
  nmap ,f <Plug>(fz)
  let g:fz_command = 'fzf'
  let g:fz_command_files = ''
  let g:fz_command_options_action = '--expect=%s'
  let g:ctrlp_map = ''
  nnoremap <C-p> :call fz#run({ 'type': 'cmd', 'cmd': 'git ls-files' })<CR>
else
```
Feel free to send a PR support for extending vim-lsp something like this.

```vim
call lsp#register_list_ui(myfzflistui#options())
let g:lsp_list_ui = ['fzf', 'quickfix', 'loclist']`
```

mkwork:
Fuzzy is not particular solution. That's algorithm.

jiz4oh:
Hi everyone, I have created a plugin [vim-lspfuzzy](https://github.com/jiz4oh/vim-lspfuzzy) that can show etc. documentsymbol and references in fzf window, any suggestion is appreciated and pr welcome

PR #226, opened 2018-12-28
install python3-dev for linux

prabirshrestha:
hopefully no build errors when running pip in linux


prabirshrestha:
didnt' work. seems like some other issue

PR #237, opened 2019-01-10
Add event queue

mattn:


mattn:
Merged @soywod's [change](https://github.com/prabirshrestha/vim-lsp/pull/188/) into this.

autoload/lsp.vim
```diff
@@ -148,28 +149,48 @@ function! s:unregister_events() abort
...
 endfunction
 
 function! s:on_text_document_did_save() abort
-    call lsp#log('s:on_text_document_did_save()', bufnr('%'))
     let l:buf = bufnr('%')
-    for l:server_name in lsp#get_whitelisted_servers()
-        call s:ensure_flush(bufnr('%'), l:server_name, {result->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})
-    endfor
+    call lsp#log('s:on_text_document_did_save()', l:buf)
+    call s:add_event_queue([l:buf, {result->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))}])
```
mattn: l:server_name is undefined

autoload/lsp.vim
```diff
@@ -148,28 +149,48 @@ function! s:unregister_events() abort
     doautocmd User lsp_unregister_server
 endfunction
 
+let s:event_queue = []
```
prabirshrestha: Since vim doesn't have concept of private only to plugin, I added most of the private functions in same file.
Should we start moving to different file so it is easier to write tests as well maintain. We could prefix private functions with `_`

mattn: Where should we move this? `lsp#utils` ?

prabirshrestha:
This isn't working for me.

image

autoload/lsp.vim
prabirshrestha: Seems like it can be a full different file `lsp#utils#_event_queue#add()`

mattn:
Seems to be related to https://github.com/prabirshrestha/vim-lsp/pull/238

prabirshrestha:
Yes. That fixed it.

autoload/lsp.vim
mattn: Hmm, ensure_flush is in script scope of lsp.vim . Seems not easier.

prabirshrestha: Makes sense.

mattn:
BTW, current implementation does not use range paramter in didChangeDocumentNotification. I know this is difficualt that vim does not provide way to know where the part of text is modified. So vim-lsp must send whole content into the LSP server. This is much heavy while editing. Do you have something idea for this?

prabirshrestha:
Here were two solutions I had thought of before.

1. Have a buffer manager that tracks old flushed buffer i.e. save all the lines as array just like vim. Then just before sending flush, do string check and diff on our own and send only text edits that actually changed. This works for all vim.
2. I believe vim now embeds a diff library instead of spawning a process for diff. Expose a function in vim that allows us to use it `let diff = diff(text1, text2)`. This still needs a bufmanager on our side. Then use the diff to create text edits. We might need to variants of `diff` in vim - one for the normal + - diffs and the other one returns proper objects so we don't have to parse the diff again. Then we just convert the parsed object to lsp text edits. 

2 might be cheaper on our side if vim supports `diff` as a function. And I'm ok only supportting that. Because vim-lsp is used by devs in their machine and mostly not in servers, it is ok to only support latest vim and tell the user to upgrade. Another example is the current diagnostics code we should probably remove and only support the new signs apis that were recently added. Makes the code smaller and easier to maintain and no hacks on our side for perf. Basically if vim is the right place for the patch, we should do it there.

I also wished if vim had some sort of webworkers concept that would allow us to do expensive computations on a different thread such as diff and fuzzy matches. 
* https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers  * https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Transferring_data_to_and_from_workers_further_details

```vim
" main.vim
let worker = start_worker('worker.vim', { type: 'file', 'message': {m->echo json_encode(m)})
call post_message(worker, {'hello': 'world'})

" worker.vim
au WorkerMessage call <sid>message()

function s:message()
    call post_message(v:message.id, v:message.data)
endfunction
```

If type is `script` then we can inline script instead of passing in file path.

Or vim can also ship with lua by default like neovim or at least have module exposed in lua by default that has thread apis. https://github.com/mah0x211/lua-pthread. If lua with threading is supported we will do expensive computation there else we will do using vimscript. This is why vim-lsp has `ui/vim` folder so that in future we could have something like `ui/lua`. apply edits can still be used written in vim so that it can be used by either vim/lua. Unless vim ships with lua by default it doesn't make sense to write plugins in lua. https://github.com/vim/vim/issues/2115

autoload/lsp.vim
prabirshrestha: we should document what object `s:event_queue` can store.

autoload/lsp.vim
```diff
@@ -148,28 +149,48 @@ function! s:unregister_events() abort
...
 endfunction
 
 function! s:on_text_document_did_save() abort
-    call lsp#log('s:on_text_document_did_save()', bufnr('%'))
     let l:buf = bufnr('%')
-    for l:server_name in lsp#get_whitelisted_servers()
-        call s:ensure_flush(bufnr('%'), l:server_name, {result->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})
-    endfor
+    call lsp#log('s:on_text_document_did_save()', l:buf)
+    call s:add_event_queue([l:buf, {server_name, result->s:call_did_save(l:buf, server_name, result, function('s:Noop'))}])
```
prabirshrestha: What happens if the event queue callback is called but the buffer is already removed?

mattn:
Please wait merge since I'm considering whether didOpen/didSave send queue immediately.

autoload/lsp.vim
```diff
@@ -697,3 +698,29 @@ endfunction
 function! lsp#complete(...) abort
     return call('lsp#omni#complete', a:000)
 endfunction
+
+let s:event_queue = []
+let s:event_timer = -1
+
+function! lsp#add_event_queue(queue) abort
```
prabirshrestha: Any particular reason `lsp#add_event_queue` is public? I find this internal implementation of vim-lsp so would be better to make it private if possible.

autoload/lsp.vim
```diff
@@ -697,3 +698,29 @@ endfunction
...
+        endif
+    endfor
+    call add(s:event_queue, a:queue)
+    call lsp#log('s:send_event_queue() will be triggered')
+    call timer_stop(s:event_timer)
+    let lazy = &updatetime > 1000 ? &updatetime : 1000
+    let s:event_timer = timer_start(lazy, function('s:send_event_queue'))
+endfunction
+
+function! s:send_event_queue(timer) abort
```
prabirshrestha: I remember in some version of neovim `timer` as arg was not present. One way to get around it would be to use `function! s:send_event_queue(...) abort` instead.

autoload/lsp.vim
```diff
@@ -654,16 +654,17 @@ function! lsp#get_whitelisted_servers(...) abort
     return l:active_servers
 endfunction
 
-function! s:get_text_document_text(buf) abort
-    return join(getbufline(a:buf, 1, '$'), "\n")
+function! s:get_text_document_text(server_name, buf) abort
+    return join(s:get_last_file_content(a:server_name, a:buf), "\n")
```
prabirshrestha: LSP protocol itself can handle `\n` or `\r\n` but do we want to explicitly send the correct line endings? Might be have a func in util that has `lsp#utils#line_ending(bufnr)`?

Also seems like we have removed. `s:requires_eol_at_eof`. More discussion about it can be found at https://github.com/prabirshrestha/vim-lsp/pull/158. It actually breaks pyls.

prabirshrestha:
`didOpen` I think makes sense to send it without queuing. `didSave` is arguable specially as some people could be calling it multiple times fast. But should be easy to change either way.

autoload/lsp.vim
mattn: If my understanding is correct, it's not required that the file content is same as server's one.

mattn:
Now works fine for me.

autoload/lsp.vim
```diff
@@ -703,3 +704,31 @@ endfunction
...
+let s:didchange_queue = []
+let s:didchange_timer = -1
+
+function! s:add_didchange_queue(buf) abort
+    if index(s:didchange_queue, a:buf) != -1
+        return
+	endif
+    call add(s:didchange_queue, a:buf)
+    call lsp#log('s:send_didchange_queue() will be triggered')
+    if exists('*timer_start')
```
prabirshrestha: We don't need to check for `timer_start`, vim-lsp requires lambda and timers to work properly.

Or did you mean `s:didchange_timer` and `s:didchange_timer > 0`

autoload/lsp.vim
```diff
@@ -150,6 +150,7 @@ function! s:register_events() abort
         autocmd BufWinLeave * call s:on_text_document_did_close()
         autocmd BufWipeout * call s:on_buf_wipeout(bufnr('<afile>'))
         autocmd InsertLeave * call s:on_text_document_did_change()
+        autocmd TextChanged * call s:on_text_document_did_change()
```
prabirshrestha: It would be good to also have `TextChangedP`, but we need to make sure it exists using `exists('##TextChangedP')`

https://github.com/vim/vim/issues/1691
https://github.com/vim/vim/pull/2372

autoload/lsp.vim
mattn: Ah, sorry. I was confused. Will fix.

prabirshrestha:
@mattn Is this ready? Seems like has merge conflicts.

mattn:
Yes, okay to merge.

>Seems like has merge conflicts.

Really?

prabirshrestha:
Yeah. Merged! Thanks a lot for this PR. Will help in perf.

Seems like my github ui defaulted in rebase and hence that had conflicts. I selected merge commit and it works.

soywod:
I see you merged [my PR](https://github.com/prabirshrestha/vim-lsp/pull/188/) in yours (thank you btw :smiley:), but it's still not working. When I use normal commands (like `dd`, `x`), the ui is not refreshing, I still have error showing.

lsp

prabirshrestha:
It is disabled by default due to some bugs. Have you enabled the flag?

mattn:
Please try:

```
let g:lsp_use_event_queue = 1
```

soywod:
Does not change anything :/

soywod:
It's even worst. Do you want me to setup a tmp repo to reproduce?

mattn:
Could you please paste log?

```
let g:lsp_log_verbose=1
let g:lsp_log_file='/tmp/lsp.log'
```

soywod:
```
Wed 23 Jan 2019 10:46:57 CET:["<---", 20, "typescript-language-server", {"response": {"method": "textDocument/publishDiagnostics", "jsonrpc": "2.0", "params": {"uri": "file:///xxx/component/", "diagnostics": [{"source": "typescript", "range": {"end": {"character": 78, "line": 0}, "start": {"character": 78, "line": 0}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 77, "line": 1}, "start": {"character": 77, "line": 1}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 59, "line": 2}, "start": {"character": 59, "line": 2}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 23, "line": 3}, "start": {"character": 23, "line": 3}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 111, "line": 4}, "start": {"character": 111, "line": 4}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 80, "line": 5}, "start": {"character": 80, "line": 5}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 80, "line": 6}, "start": {"character": 80, "line": 6}}, "code": 1002, "message": "Unterminated string literal.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 2, "line": 7}, "start": {"character": 1, "line": 7}}, "code": 1003, "message": "Identifier expected.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 3, "line": 7}, "start": {"character": 2, "line": 7}}, "code": 1003, "message": "Identifier expected.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 1, "line": 8}, "start": {"character": 0, "line": 8}}, "code": 1109, "message": "Expression expected.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 2, "line": 8}, "start": {"character": 1, "line": 8}}, "code": 1003, "message": "Identifier expected.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 5, "line": 10}, "start": {"character": 5, "line": 10}}, "code": 1109, "message": "Expression expected.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 3, "line": 9}, "start": {"character": 0, "line": 9}}, "code": 2304, "message": "Cannot find name 'App'.", "severity": 1}, {"source": "typescript", "range": {"end": {"character": 4, "line": 10}, "start": {"character": 0, "line": 10}}, "code": 2304, "message": "Cannot find name 'Card'.", "severity": 1}]}}}]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:46:57 CET:["add signs"]
Wed 23 Jan 2019 10:47:05 CET:["s:on_text_document_did_close()", 5]
```

mattn:
Sorry, now I noticed what you expect. Do you expect that sign are updated while editing without waiting few seconds. Right? Then:

Please try to add the following code in s:register_events in autoload/lsp.vim.

```
autocmd TextChangedI * call s:on_text_document_did_change()
```

soywod:
Not really. Let me explain:
- When I enter in `insert` mode, insert char, then exit, the sign is correctly shown (both this repo and my fork).
- When I perform a `normal` action like `dd`, `x`, the sign does not update, and the error message is still there (but only on this repo, it works well with [my fork](https://github.com/soywod/vim-lsp/tree/text-changed-autocmd)).

It's this last one I tried to fix with my PR, and that seems not to work even with your PR.

mattn:
If you use this change, the log file above should contains `s:on_text_document_did_change`. But seems not.

mattn:
As far as I can see the log, your environment does not fire the timer. What do you set value for `updatetime` option?

soywod:
My fork doesn't have your timer version, it's normal not to see it. My `updatetime` = `4000`.

mattn:
@prabirshrestha I figure out your https://github.com/prabirshrestha/vim-lsp/pull/237#issuecomment-453326600 .

```
function! lsp#send_request(server_name, request) abort
    let l:Cb = has_key(a:request, 'on_notification') ? a:request['on_notification'] : function('s:Noop')
    let l:request = copy(a:request)
    let l:request['on_notification'] = {id, data, event->l:Cb(data)}
    call lsp#utils#step#start([
        \ {s->s:ensure_flush(bufnr('%'), a:server_name, s.callback)},
        \ {s->s:is_step_error(s) ? l:Cb(s.result[0]) : s:send_request(a:server_name, l:request) },
        \ ])
endfunction
```
Current implementation, use `%` for the buffer, but current buffer might be changed with tag jump, or switching buffer, showing help, etc. So I suggest to change this function to add bufnr, or add new function which take argument bufnr.

mattn:
Or third optional argument?

prabirshrestha:
Optional third argument seems good.

We could add deprecation echo if people pass only 2 args. We would need to update asyncomplete vim-lsp.

mattn:
Ah, sorry. It is NOT need to fix `lsp#send_request`.

```
    for l:server in l:servers
        call lsp#send_request(l:server, {
            \ 'method': 'textDocument/definition',
            \ 'params': {
            \   'textDocument': lsp#get_text_document_identifier(),
            \   'position': lsp#get_position(),
            \ },
            \ 'on_notification': function('s:handle_location', [l:ctx, l:server, 'definition']),
            \ })
    endfor
```
Should fix `s:handle_location` to take one more argument.

PR #251, opened 2019-01-16
 Fix s:has_bool_provider to be possible to look nested provider names.

mattn:
Support boolean or RenameOptions.
```
	/**
	 * The server provides rename support. RenameOptions may only be
	 * specified if the client states that it supports
	 * `prepareSupport` in its initial `initialize` request.
	 */
	renameProvider?: boolean | RenameOptions;
```


mattn:
Without this change, I got error with jdt language server. So merging.

PR #257, opened 2019-01-19
Disable event queue in default

mattn:
See https://github.com/prabirshrestha/vim-lsp/issues/256


mattn:
Perhaps, better to limit the insert-mode for about using the event queue.

PR #273, opened 2019-01-29
Fix race while switching buffer

mattn:
Using event queue. Modifying buffer and switching buffer. This stock two events in the queue. But lsp#get_whitelisted_servers return the servers for the current buffer. So vim-lsp send content to unrelated langserver. Probably, this will fix issue that @prabirshrestha got in few days.

https://github.com/prabirshrestha/vim-lsp/pull/237#issuecomment-453326600


mattn:
Confirmed lsp_use_event_queue 0/1 both.

PR #280, opened 2019-02-06
Fix race condition when switching buffer

mattn:
Fixes https://github.com/prabirshrestha/vim-lsp/issues/256


mattn:
wrong approach.

PR #298, opened 2019-02-10
Fixes exception for empty list

mattn:


mattn:
Repro step

1. set lsp_use_event_queue = 1
2. open a file
3. modify and undo immediately
4. tag jump within 4 seconds (4 is updatetime)

PR #306, opened 2019-02-17
UltiSnips integration

thomasfaingnaert:
I added a new variable `g:lsp_ultisnips_integration` which, when set to `1`, uses UltiSnips to provide LSP snippet functionality.
If `g:lsp_ultisnips_integration = 0` (the default), vim-lsp should continue to function in exactly the same way as it does now, so this PR should not break anything.
This requires UltiSnips to be installed, and Vim 8.0 patch 1493 or higher (for `user_data`).

A couple notes:
* Because UltiSnips deletes all tabstops when switching buffers, you have to remove `preview` from `'completeopt'`. Alternatively, you can try https://github.com/SirVer/ultisnips/pull/1055.
* If you want snippets to expand when its trigger is the only match, you need to add `menuone` to `'completeopt'`. This is because in that case, the `CompleteDone` autocommand is called with `v:completed_item = {}` instead of the completed item. If anyone knows another fix for this, please let me know.
* UltiSnips does not implement the complete LSP spec. For example, choice placeholders will just be displayed verbatim. However, because I am most familiar with UltiSnips as a vim snippet plugin, I decided to use it instead of other solutions. Besides, most of the important features are present.
* I tested this with omnifunc, however because I am using `CompleteDone`, I believe this should also work with asyncomplete.vim.

Demo with C++ in clangd:
demo

Minimal vimrc to reproduce:
```vim
call plug#begin()

Plug 'SirVer/ultisnips'
Plug 'thomasfaingnaert/vim-lsp', {'branch': 'ultisnips-integration'}

call plug#end()

let g:UltiSnipsExpandTrigger="<tab>"
let g:UltiSnipsJumpForwardTrigger="<tab>"
let g:UltiSnipsJumpBackwardTrigger="<s-tab>"

if executable('clangd')
    augroup vim_lsp_cpp
        autocmd!
                    \ 'name': 'clangd',
                    \ 'cmd': {server_info->['clangd']},
                    \ 'whitelist': ['c', 'cpp', 'objc', 'objcpp', 'cc'],
                    \ })
        autocmd FileType c,cpp,objc,objcpp,cc setlocal omnifunc=lsp#complete
    augroup end
endif

set completeopt-=preview
set completeopt+=menuone

let g:lsp_ultisnips_integration = 1
```

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -178,12 +178,6 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
         let l:abbr = a:item['label']
     endif
 
-    if g:lsp_ultisnips_integration && has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] == 2
-        let l:snippet = substitute(a:item['insertText'], '\%x00', '\\n', 'g')
-        let l:word = trim(a:item['label'])
-        let l:trigger = l:word
-    endif
-
     if l:do_remove_typed_part
         let l:word = s:remove_typed_part(l:word)
     endif
@@ -205,10 +199,6 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
         if type(a:item['documentation']) == type('')
             let l:completion['info'] .= a:item['documentation']
         endif
-    endif
-
-    if exists('l:snippet')
-        let l:completion['user_data'] = string([l:trigger, l:snippet])
     endif
 
     return l:completion
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -212,11 +212,11 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
     return l:completion
 endfunction
 
-function! s:expand_snippet(timer)
+function! s:expand_snippet(timer) abort
     call feedkeys("\<C-r>=UltiSnips#Anon(\"" . s:snippet . "\", \"" . s:trigger . "\", '', 'i')\<CR>")
 endfunction
 
-function! s:handle_snippet(item)
+function! s:handle_snippet(item) abort
     if !has_key(a:item, 'user_data')
         return
     endif
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -235,7 +235,7 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
 endfunction
 
 function! lsp#omni#get_vim_completion_item(...) abort
-    return call(g:Lsp_get_vim_completion_item, a:000)
+    return call(g:lsp_get_vim_completion_item[0], a:000)
 endfunction
 
 augroup lsp_completion_item_text_edit
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -269,16 +269,19 @@ function! s:apply_text_edits() abort
     "       ],
     "     }
     if !g:lsp_text_edit_enabled
+        doautocmd User lsp_complete_done
         return
     endif
 
     " completion faild or not select complete item
     if empty(v:completed_item)
+        doautocmd User lsp_complete_done
         return
     endif
 
     " check user_data
     if !has_key(v:completed_item, 'user_data')
+        doautocmd User lsp_complete_done
         return
     endif
 
@@ -287,10 +290,12 @@ function! s:apply_text_edits() abort
         let l:user_data = json_decode(v:completed_item['user_data'])
     catch
         " do nothing if user_data is not json type string.
+        doautocmd User lsp_complete_done
         return
     endtry
 
     if type(l:user_data) != type({})
+        doautocmd User lsp_complete_done
         return
     endif
 
@@ -323,6 +328,8 @@ function! s:apply_text_edits() abort
     let l:pos[2] += l:col_offset
     call setpos("'a", l:saved_mark)
     call setpos('.', l:pos)
+
+    doautocmd User lsp_complete_done
 endfunction
 
 function! s:expand_range(text_edit, expand_length) abort
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -212,6 +212,10 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
     return l:completion
 endfunction
 
+function! s:expand_snippet(timer)
+    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . s:snippet . "\", \"" . s:trigger . "\", '', 'i')\<CR>")
+endfunction
+
 function! s:handle_snippet(item)
     if !has_key(a:item, 'user_data')
         return
@@ -219,10 +223,10 @@ function! s:handle_snippet(item)
 
     execute 'let l:user_data = ' . a:item['user_data']
 
-    let l:trigger = l:user_data[0]
-    let l:snippet = l:user_data[1]
+    let s:trigger = l:user_data[0]
+    let s:snippet = l:user_data[1]
 
-    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . l:snippet . "\", \"" . l:trigger . "\", '', 'i')\<CR>")
+    call timer_start(0, function('s:expand_snippet'))
 endfunction
 
 if g:lsp_ultisnips_integration
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -180,6 +180,7 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
 
     if g:lsp_ultisnips_integration && has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] == 2
         let l:snippet = substitute(a:item['insertText'], '\%x00', '\\n', 'g')
+        let l:word = trim(a:item['label'])
         let l:trigger = l:word
     endif
 
@@ -218,7 +219,7 @@ function! s:handle_snippet(item)
 
     execute 'let l:user_data = ' . a:item['user_data']
 
-    let l:trigger = trim(l:user_data[0])
+    let l:trigger = l:user_data[0]
     let l:snippet = l:user_data[1]
 
     call feedkeys("\<C-r>=UltiSnips#Anon(\"" . l:snippet . "\", \"" . l:trigger . "\", '', 'i')\<CR>")
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -178,6 +178,11 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
         let l:abbr = a:item['label']
     endif
 
+    if g:lsp_ultisnips_integration && has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] == 2
+        let l:snippet = substitute(a:item['insertText'], '\%x00', '\\n', 'g')
+        let l:trigger = l:word
+    endif
+
     if a:do_remove_typed_part
         let l:word = s:remove_typed_part(l:word)
     endif
@@ -199,8 +204,8 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
         endif
     endif
 
-    if g:lsp_ultisnips_integration && has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] == 2
-        let l:completion['user_data'] = a:item['insertText']
+    if exists('l:snippet')
+        let l:completion['user_data'] = string([l:trigger, l:snippet])
     endif
 
     return l:completion
@@ -211,10 +216,12 @@ function! s:handle_snippet(item)
         return
     endif
 
-    let l:snippet = substitute(a:item['user_data'], '\%x00', '\\n', 'g')
-    let s:trigger = a:item['word']
+    execute 'let l:user_data = ' . a:item['user_data']
 
-    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . l:snippet . "\", \"" . s:trigger . "\", '', 'i')\<CR>")
+    let l:trigger = trim(l:user_data[0])
+    let l:snippet = l:user_data[1]
+
+    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . l:snippet . "\", \"" . l:trigger . "\", '', 'i')\<CR>")
 endfunction
 
 if g:lsp_ultisnips_integration
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -199,7 +199,29 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
         endif
     endif
 
+    if g:lsp_ultisnips_integration && has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] == 2
+        let l:completion['user_data'] = a:item['insertText']
+    endif
+
     return l:completion
 endfunction
+
+function! s:handle_snippet(item)
+    if !has_key(a:item, 'user_data')
+        return
+    endif
+
+    let l:snippet = substitute(a:item['user_data'], '\%x00', '\\n', 'g')
+    let s:trigger = a:item['word']
+
+    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . l:snippet . "\", \"" . s:trigger . "\", '', 'i')\<CR>")
+endfunction
+
+if g:lsp_ultisnips_integration
+    augroup lsp_ultisnips
+        autocmd!
+        autocmd CompleteDone * call s:handle_snippet(v:completed_item)
+    augroup end
+endif
 
 " }}}
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -208,28 +208,4 @@ function! lsp#omni#get_vim_completion_item(...) abort
     return call(g:Lsp_get_vim_completion_item, a:000)
 endfunction
 
-function! s:expand_snippet(timer) abort
-    call feedkeys("\<C-r>=UltiSnips#Anon(\"" . s:snippet . "\", \"" . s:trigger . "\", '', 'i')\<CR>")
-endfunction
-
-function! s:handle_snippet(item) abort
-    if !has_key(a:item, 'user_data')
-        return
-    endif
-
-    execute 'let l:user_data = ' . a:item['user_data']
-
-    let s:trigger = l:user_data[0]
-    let s:snippet = l:user_data[1]
-
-    call timer_start(0, function('s:expand_snippet'))
-endfunction
-
-if g:lsp_ultisnips_integration
-    augroup lsp_ultisnips
-        autocmd!
-        autocmd CompleteDone * call s:handle_snippet(v:completed_item)
-    augroup end
-endif
-
 " }}}
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -163,7 +163,7 @@ function! s:remove_typed_part(word) abort
     return strpart(a:word, l:overlap_length)
 endfunction
 
-function! lsp#omni#get_vim_completion_item(item, ...) abort
+function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     let l:do_remove_typed_part = get(a:, 1, 0)
 
     if g:lsp_insert_text_enabled && has_key(a:item, 'insertText') && !empty(a:item['insertText'])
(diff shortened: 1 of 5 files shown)


thomasfaingnaert:
Concerning the failing CI build: the issue is this snippet:
```vim
execute 'let l:user_data = ' . a:item['user_data']

let s:trigger = l:user_data[0]
let s:snippet = l:user_data[1]
```
which produces `autoload/lsp/omni.vim:228:21: Undefined variable: l:user_data (see :help E738)`.

This snippet is used to convert a string back to an array, which was needed because `user_data` must be a string.
One way to fix this would be to add an explicit assignment `let l:user_data = ''` before the `execute`, but perhaps there is a better way to do the conversion from string to array?

prabirshrestha:
Because none of the vim snippets actually support lsp I'm a bit hesitant to support this out of the box. Similar to how autocomplete works with asyncomplete.vim and deplolete, we can easily make this work with an external plugin.

1. Expose a function for to override capabilities so someone outside vim-lsp can set it.
2. Allow `lsp#omni#get_vim_completion_item` to be overriden so someone outside vim-lsp can do whatever they want. They can manually set `CompleteDone` to handle snippet trigger.

This allows to have external plugins such as `vim-lsp-ultisnips`, `vim-lsp-neosnippet` and so on.

autoload/lsp/omni.vim
```diff
@@ -201,7 +207,35 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
         endif
     endif
 
+    if exists('l:snippet')
+        let l:completion['user_data'] = string([l:trigger, l:snippet])
```
prabirshrestha: `user_data` is only available in certain version of vim. We need to check if vim has that patch.

autoload/lsp/omni.vim
```diff
@@ -201,7 +207,35 @@ function! lsp#omni#get_vim_completion_item(item, ...) abort
...
+    if !has_key(a:item, 'user_data')
+        return
+    endif
+
+    execute 'let l:user_data = ' . a:item['user_data']
+
+    let s:trigger = l:user_data[0]
+    let s:snippet = l:user_data[1]
+
+    call timer_start(0, function('s:expand_snippet'))
```
prabirshrestha: Curious why do we need a timer here?

autoload/lsp/omni.vim
thomasfaingnaert: In `plugin/lsp.vim`, I already check if Vim `has('patch-8.0.1493')`. Would you prefer I set `g:lsp_ultisnips_integration` back to `0` there and/or add an additional `has('patch-8.0.1493')` to `if exists('l:snippet')`?

autoload/lsp/omni.vim
thomasfaingnaert: I'm not 100% sure why, but without a timer the tabstops in nested snippets do not work:
without-timer

It seems only the innermost snippet's tabstops are preserved.

thomasfaingnaert:
I agree, perhaps it would be a better idea to separate the snippet plugin dependent logic in a separate plugin, and allow that plugin to change the behaviour of `vim-lsp` using an API.
How do you imagine such an interface might look like?

prabirshrestha:
My thought was to make it simple as this.

```vim
let g:lsp_get_vim_completion_item = get(g:, 'myplugin#utilsnips#get_vim_completion_item', 'lsp#omni#get_vim_completion_item')
```

We might want to rename existing `lsp#omni#get_vim_completion_item` to `lsp#omni#get_default_vim_completion_item` so `get_vim_completion_item` can actually call `g:lsp_get_vim_completion_item` internally. Note: we do need to make sure we don't break asyncomplete-lsp.vim.

thomasfaingnaert:
So if I understand correctly, you're proposing the following:

In `vim-lsp`:
- Move all existing logic to `lsp#omni#get_default_vim_completion_item(item, ...)`
- Introduce a global variable `g:lsp_get_vim_completion_item` that points to the completion item provider function that should be used
- Change `lsp#omni#get_vim_completion_item(item, ...)` to just forward its arguments to a call to `g:lsp_get_vim_completion_item`
- Add `let g:lsp_get_vim_completion_item = get(g:, 'myplugin#ultisnips#get_vim_completion_item', 'lsp#omni#get_default_vim_completion_item')` to `vim-lsp` for all snippet plugin integrations
- Add snippet support to the server's capabilities, e.g. depending on the value of a boolean variable `g:lsp_enable_snippets`

In `vim-lsp-ultisnips`:
- Let `myplugin#ultisnips#get_vim_completion_item(item, ...)` call `lsp#omni#get_default_vim_completion_item(item, ...)`, and change the `user_data` of its return value
- Register a `CompleteDone` autocmd to expand the snippet

If so, would it not be better to just set `g:lsp_get_vim_completion_item` to `lsp#omni#get_default_vim_completion_item` in `vim-lsp`, and let `vim-lsp-ultisnips` change `g:lsp_get_vim_completion_item` to its own provider `myplugin#...` when it is loaded? That way, writing a new snippet plugin integration does not require changing `vim-lsp`'s source.

prabirshrestha:
That is true. It is possible but there is quite a logic that could be reused by others from `get_default_vim_completion_item`, unless you want to copy paste the code.

prabirshrestha:
Instead of `g:lsp_enable_snippets` might be wroth adding that as function too so people can add other custom capabilities but that does mean a bit more work but a snippet plugin can handle it so others don't have to worry about it,

natebosch:
Some snippet managers, like `neosnippet`, allow using `CompleteDone` and reading from `user_data` on a completed item in order to expand a snippet. I don't know if `UltiSnips` has this built in but I'd imagine it would be feasible to build something similar. You might consider going this route so that it's up to the snipper plugin to handle and read `user_data` however it wants, and the implementation here is not tied to any one plugin.

You can see an example of this approach and some discussion here: https://github.com/natebosch/vim-lsc/pull/142

thomasfaingnaert:
@natebosch Thanks for the info! The way I'm rewriting it now is to allow external plugins to provide their own implementation of `get_vim_completion_item` and to allow them to specify the client capabilities. That way, an external plugin can customize `get_vim_completion_item` to set `user_data` to whatever it desires, add snippets to the capabilities, and register its own `CompleteDone` autocmd.
Of course, the custom versions of these functions will be able to call the default `vim-lsp` version of these functions (to avoid code duplication).
I think this would be the most general solution to provide snippet integration.

prabirshrestha:
I would still prefer func instead of setting `snippet` in `user_data` as the snippet format is not defined.
Asyncomplete supports other completion besides LSP so we need a way to differentiate it.

thomasfaingnaert:
@prabirshrestha Is this what you had in mind?
Example usage:
```vim
function! s:custom_get_item(item, ...) abort
        let l:completion = call(function('lsp#omni#default_get_vim_completion_item'), [a:item] + a:000)
        " let l:completion['user_data'] = ...
       return l:completion
endfunction

function! s:custom_get_caps() abort
        let l:caps = lsp#default_get_supported_capabilities()
        " add snippets to capabilities etc...
       return l:caps
endfunction

let g:Lsp_get_vim_completion_item = function('s:custom_get_item')
let g:Lsp_get_supported_capabilities = function('s:custom_get_caps')
```

thomasfaingnaert:
Using this PR, I wrote two plugins to add LSP snippets in vim. One using UltiSnips ([vim-lsp-ultisnips](https://github.com/thomasfaingnaert/vim-lsp-ultisnips)) and one using neosnippet.vim ([vim-lsp-neosnippet](https://github.com/thomasfaingnaert/vim-lsp-neosnippet)). Would it be OK to add a reference to these two repositories in `vim-lsp`'s README?

prabirshrestha:
Feel free to reference it in README.md as well as in docs.

mattn:
@prabirshrestha I feed this is good to merge. Any thought?

prabirshrestha:
I think it should be good but even better if we actually passed server info as params to the func.

```
let l:capabilities = call(g:Lsp_get_supported_capabilities, [server_info])
```

Then we can do something like this if we want to disable snippets per server.

```
    \ 'name': 'pyls',
    \ 'whitelist': ['python'],
    \ 'config': { 'utilsnips': 0 }
    \ })
```

In asyncomplete.vim I reserved the config for the sources. https://github.com/prabirshrestha/asyncomplete-tags.vim

```
au User asyncomplete_setup call asyncomplete#register_source(asyncomplete#sources#tags#get_source_options({
    \ 'name': 'tags',
    \ 'whitelist': ['c'],
    \ 'completor': function('asyncomplete#sources#tags#completor'),
    \ 'config': {
    \    'max_file_size': 50000000,
    \  },
    \ }))
```

thomasfaingnaert:
@mattn I just need to add some documentation, then this is ready I feel.
@prabirshrestha That's a good idea, I'll see if I can implement this as well.

thomasfaingnaert:
I added info to readme and docs about snippets and the new configuration options.
It is now also possible to disable vim-lsp-ultisnips and vim-lsp-neosnippet using `'config': { 'snippets': 0 }`.
I'll fix the conflict in `doc/vim-lsp.txt`, and then I think this is ready to merge.

prabirshrestha:
@thomasfaingnaert @mattn I think `textEdit` support should always be enabled for rename/codeactions to work. It is the `completion.completionItem.snippetSupport` that needs to be set to true. Let me know if I'm miss-reading the sepc.

thomasfaingnaert:
@prabirshrestha I'm not entirely sure what you mean exactly. I don't change `textEdit` support, and I set `textDocument.completion.completionItem.snippetSupport` to true in both vim-lsp-ultisnips and vim-lsp-neosnippet.

thomasfaingnaert:
@prabirshrestha Any update on this?

prabirshrestha:
@thomasfaingnaert Seems to be conflicting a bit with this PR to enable textEdits since both use user data and seems to be conflicting. https://github.com/prabirshrestha/vim-lsp/pull/318

I do want both the PRs to be merged and enable both use case. Looking at both PRs we might have to merge the text edit PR first since it doesn't expose dependencies to others.

//cc @mikoto2000

prabirshrestha:
Now that TextEdit for omnicomplete is supported can you try this against the latest master. https://github.com/prabirshrestha/vim-lsp/pull/318

plugin/lsp.vim
```diff
@@ -21,6 +21,9 @@ let g:lsp_preview_keep_focus = get(g:, 'lsp_preview_keep_focus', 1)
 let g:lsp_use_event_queue = get(g:, 'lsp_use_event_queue', has('nvim') || has('patch-8.1.0889'))
 let g:lsp_insert_text_enabled= get(g:, 'lsp_insert_text_enabled', 1)
 
+let g:Lsp_get_vim_completion_item = get(g:, 'Lsp_get_vim_completion_item', function('lsp#omni#default_get_vim_completion_item'))
```
prabirshrestha: Can we change this to be an array instead even though we don't support multiple. This is what I do in asyncomplte v2. That way we don't need to worry about the capital L.

```
let g:lsp_get_vim_completion_item = [function('sp#omni#default_get_vim_completion_item')]
```

prabirshrestha:
@thomasfaingnaert  Do you have a minimal vimrc as well as a langserver that I can try.

thomasfaingnaert:
@prabirshrestha I've mainly been using clangd, which you can install from http://releases.llvm.org/download.html or from your package manager if you're on *NIX (`clang-7` and `clang-tools-7` on Debian based distros).

<details>
<summary>Minimal vimrc</summary>

```vim
call plug#begin()

Plug 'SirVer/ultisnips'
Plug 'thomasfaingnaert/vim-lsp', {'branch': 'ultisnips-integration'}
Plug 'thomasfaingnaert/vim-lsp-ultisnips'

call plug#end()

let g:UltiSnipsExpandTrigger="<tab>"
let g:UltiSnipsJumpForwardTrigger="<tab>"
let g:UltiSnipsJumpBackwardTrigger="<s-tab>"

if executable('clangd')
    augroup vim_lsp_cpp
        autocmd!
                    \ 'name': 'clangd',
                    \ 'cmd': {server_info->['clangd']},
                    \ 'whitelist': ['c', 'cpp', 'objc', 'objcpp', 'cc'],
                    \ })
	autocmd FileType c,cpp,objc,objcpp,cc setlocal omnifunc=lsp#complete
    augroup end
endif

set completeopt+=menuone
```
</details>

<details>
<summary>Example test.cpp</summary>

```vim
#include <iostream>
#include <vector>

int main(int argc, char **argv)
{
	// type: std::co<C-x><C-o>
	// to get: std::cout
	
	// type: std::vec<C-x><C-o><C-y>int<Tab> v;
	// to get: std::vector<int> v;

	// type: vect.pu<C-x><C-o><C-y>30<Tab>;
	// to get: vect.push_back(30);
	std::vector<int> vect;

	// type: for<C-x><C-o><C-y>int i = 0<Tab>i<10<Tab>i++<Tab>// ...
	// to get: for(int i = 0;i<10;i++){
	// //...
	// }
}
```
</details>

prabirshrestha:
Thanks for the instructions. I will be out on vacation for  2 weeks, will give it a try when I come back.

mikoto2000:
What do you think adding snippet plugin integration support API?
I think more easily implement snippet plugin integration if this API added.

For example:
- https://github.com/mikoto2000/vim-lsp/commit/dfbf3f8550c66b797e2b862bba2db7c61992ed4b
- https://github.com/mikoto2000/vim-lsp-ultisnips/commit/aeeb5a4685326b9dcbc0dc9e106099f63fbbeb1e

<details>
<summary>Difficulty 2 points of snippet integration</summary>

#### 1. can not call snippet function after `s:apply_text_edits` 

I want call `search_first_placeholder_and_move` function after `s:apply_text_edits`(called by CompleteDone).
But `autocmd CompleteDone * call <SID>apply_text_edits()` is autoload, so I can not register easily `search_first_placeholder_and_move` after `s:apply_text_edits`.

https://github.com/prabirshrestha/vim-lsp/blob/185aee6ac905a03980333a413f72a5131c89d047/autoload/lsp/omni.vim#L238-L241
e.g. `:augroup` of vim-lsp and vim-lsp-ultisnips.
image


#### 2. `v:completed_item` is empty, after call `lsp#utils#text_edit#apply_text_edits`. 

`v:completed_item` is empty after `lsp#utils#text_edit#apply_text_edits`, so after called CompleteDone function can not reference `v:completed_item`.
`v:completed_item` is needed for move cursor to tabstop or placeholder.



```vim
    call lsp#log(v:completed_item) " {"word": "...(snip)} (not empty)

    " apply textEdits
    if !empty(l:all_text_edits)
        call lsp#utils#text_edit#apply_text_edits(lsp#utils#get_buffer_uri(), l:all_text_edits)
    endif

    call lsp#log(v:completed_item) " {} (empty)
```


</details>

thomasfaingnaert:
@mikoto2000
Sorry for the late response. For your first issue, how about I use a User autocmd that is called at the end of `s:apply_text_edit`, so users/other plugins can register their own callbacks as well?

For the second issue: I did some research, and I think the problem is https://github.com/vim/vim/blob/8055d17388736421d875dd4933c4c93d49a2ab58/src/insexpand.c#L1512, which is called in `edit`: https://github.com/vim/vim/blob/8055d17388736421d875dd4933c4c93d49a2ab58/src/edit.c#L204.

Because the apply_text_edit function uses `c`, `i` and `a`, and these commands enter insert mode, `v:completed_item` is cleared. `x` does not have this problem because you stay in normal mode. One way to solve this is to use `p` and `P` instead of `c`, `i` and `a`, which I have implemented in https://github.com/thomasfaingnaert/vim-lsp/commit/1371a1e59fcbd6256f2a449207ddc9e2f6eb22eb.

From my limited testing, that indeed retains the value of `v:completed_item`, without breaking anything.
If it's OK with @prabirshrestha, I'll create a PR for this as well.

mikoto2000:
@thomasfaingnaert 
Thank you for your examination and comments.

>  I use a User autocmd that is called at the end of `s:apply_text_edit,` 

> use p and P instead of c, i and a,

That makes sense.
I think those fixes are smart approach.

thomasfaingnaert:
@prabirshrestha 
I have extracted duplicated code in vim-lsp-ultisnips and vim-lsp-neosnippet in a separate repo vim-lsp-snippets, which is now a dependency as well.
The new installation instructions are very similar, but I'll repeat them here:

1. Install a language server that supports snippets, e.g. `clangd` (Windows: http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe, Ubuntu: `sudo apt install -y clang-7 clang-tools-7`). For completeness `clangd --version` gives `clangd version 7.0.1 (tags/RELEASE_701/final)` on my system.

2. Install `vim-plug`: https://github.com/junegunn/vim-plug#vim

3. Use the following minimal vimrc and run `:PlugInstall`:
```viml
call plug#begin()

Plug 'SirVer/ultisnips'
Plug 'thomasfaingnaert/vim-lsp', {'branch': 'ultisnips-integration'}
Plug 'thomasfaingnaert/vim-lsp-snippets'
Plug 'thomasfaingnaert/vim-lsp-ultisnips'

call plug#end()

let g:UltiSnipsExpandTrigger="<tab>"
let g:UltiSnipsJumpForwardTrigger="<tab>"
let g:UltiSnipsJumpBackwardTrigger="<s-tab>"

if executable('clangd')
    augroup vim_lsp_cpp
        autocmd!
                    \ 'name': 'clangd',
                    \ 'cmd': {server_info->['clangd']},
                    \ 'whitelist': ['c', 'cpp', 'objc', 'objcpp', 'cc'],
                    \ })
	autocmd FileType c,cpp,objc,objcpp,cc setlocal omnifunc=lsp#complete
    augroup end
endif

set completeopt+=menuone
```

4. Create a file `test.cpp` with the following contents, and open it in Vim:
```cpp
#include <iostream>
#include <vector>

int main(int argc, char **argv)
{
	// type: std::co<C-x><C-o>
	// to get: std::cout
	
	// type: std::vec<C-x><C-o><C-y>int<Tab> v;
	// to get: std::vector<int> v;

	// type: vect.pu<C-x><C-o><C-y>30<Tab>;
	// to get: vect.push_back(30);
	std::vector<int> vect;

	// type: for<C-x><C-o><C-y>int i = 0<Tab>i<10<Tab>i++<Tab>// ...
	// to get: for(int i = 0;i<10;i++){
	// //...
	// }
}
```

I have also tested this with a language server that uses text edits (I've been using https://github.com/prabirshrestha/vim-lsp/pull/318#issuecomment-475856672), and this works as well, provided #389 is merged.

thomasfaingnaert:
@prabirshrestha 
Now that #389 is merged, you can test this with text edits as well, if you want to.
<details>
<summary>Minimal vimrc</summary>

```viml
call plug#begin()

Plug 'SirVer/ultisnips'
Plug 'thomasfaingnaert/vim-lsp', {'branch': 'ultisnips-integration'}
Plug 'thomasfaingnaert/vim-lsp-snippets'
Plug 'thomasfaingnaert/vim-lsp-ultisnips'

call plug#end()

let g:UltiSnipsExpandTrigger="<tab>"
let g:UltiSnipsJumpForwardTrigger="<tab>"
let g:UltiSnipsJumpBackwardTrigger="<s-tab>"

\ 'name': 'lsp4xml',
\ 'cmd': {server_info->[
\     'java',
\     '-noverify',
\     '-Xmx1G',
\     '-XX:+UseStringDeduplication',
\     '-Dfile.encoding=UTF-8',
\     '-jar',
\     '/path/to/lsp4xml/org.eclipse.lsp4xml/target/org.eclipse.lsp4xml-uber.jar'
\ ]},
\ 'whitelist': ['xml', 'arxml'],
\ })

autocmd FileType xml setlocal omnifunc=lsp#complete

set completeopt+=menuone
```
</details>

And using the files from https://github.com/prabirshrestha/vim-lsp/pull/318#issuecomment-475856672:
<details>
<summary>test.xsd</summary>

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <xsd:element name="books">
    <xsd:complexType>
      <xsd:sequence>
        <xsd:element name="book">
          <xsd:complexType>
            <xsd:sequence>
              <xsd:element name="title" type="xsd:string" />
              <xsd:element name="price" type="xsd:decimal" />
            </xsd:sequence>
          </xsd:complexType>
        </xsd:element>
      </xsd:sequence>
    </xsd:complexType>
  </xsd:element>
</xsd:schema>
```
</details>

<details>
<summary>test.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<books xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="./test.xsd">
	<!-- type <C-x> <C-o> and select book, <C-y> -->
</books>
```
</details>

prabirshrestha:
Merged. Thanks to all for the PR feedback as well as other fixes.

PR #309, opened 2019-02-19
Check &hidden for split window

### LINKED ISSUE #307: Add option to disable split when going to definition
A change was recently merged (#277) that causes "go to definition" to open a split when the current buffer has been edited. I, at least, much preferred the previous behavior. The change seems to have been made to deal with E37 errors, but that should only come up if `hidden` is disabled. A better solution would be to only use the new behavior if `hidden` is disabled, or to make the new behavior a configurable option.

mattn:
Fixes #307


mattn:
Thank you

PR #329, opened 2019-03-02
initializationOptions is not passed

mattn:


mattn:
@ElArtista Could you please check this?

PR #352, opened 2019-03-22
Add support for nvim highlights

jason0x43:
This PR adds support for neovim text highlights (underlining). It's organized in the same way as the recently merged virtual text PR.

I used a different highlight group than for signs and virtual text because I feel that these should generally look different. For example, a highlight group is underlined, a sign probably doesn't need an underline since it's in the sign column, and virtual text should be easily distinguishable from other text (more than just an underline or color change could give it).


prabirshrestha:
Merged. Works great. Thanks.

Screen Shot 2019-03-23 at 10 00 43 AM

PR #375, opened 2019-05-05
Add diagnostics highlighting via +textprop.

tsufeki:
Add diagnostics highlights using text properties introduced in vim patch 8.1.0579

Turned off by default as text properties are not stable yet.


prabirshrestha:
Could you provide screenshots and gifs?

tsufeki:
image

prabirshrestha:
Looks good to me. Can we enable this by default and add docs?

tsufeki:
Done. I also added docs for `g:lsp_highlights_enabled`.

Text props don't seem to be stable yet though (I occasionally get an "internal error" when editing). Should we assume users know what they're doing by using development version of vim?

prabirshrestha:
If it isn’t stable ok to not enable it by default.

prabirshrestha:
Merged. Thanks.

Once text prop gets stable we most likely can remove that flag and just use highlights flag.

PR #389, opened 2019-05-22
Fix text edits not preserving v:completed_item

thomasfaingnaert:
As discussed in https://github.com/prabirshrestha/vim-lsp/pull/306#issuecomment-491477815 and https://github.com/prabirshrestha/vim-lsp/pull/306#issuecomment-493664731, `lsp#utils#text_edit#apply_text_edits` does not preserve `v:completed_item`, because it uses `c`, `a` and `i`, and entering insert mode resets `v:completed_item`.

I fixed this by using `p` and `P` instead. I also added a test to make sure this bug remains fixed.

Slight note: the text_edit command is `execute`'d in `apply_text_edits`, however the command is built in a different function. Thus I had to use the variable `l:merged_text_edit` (local to `apply_text_edits`) inside `generate_sub_cmd_...`. I'm not a huge fan of this, so perhaps you have a better suggestion on how to tackle this?

/cc @mikoto2000

### DIFF
--- a/autoload/lsp/utils/text_edit.vim
+++ b/autoload/lsp/utils/text_edit.vim
@@ -158,12 +158,10 @@ function! s:generate_sub_cmd_insert(text_edit) abort
     let l:sub_cmd .= s:generate_move_start_cmd(l:start_line, l:start_character)
 
     if l:start_character >= len(getline(l:start_line))
-        let l:sub_cmd .= 'a'
+        let l:sub_cmd .= "\"=l:merged_text_edit['merged']['newText']\<CR>p"
     else
-        let l:sub_cmd .= 'i'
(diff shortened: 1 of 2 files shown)


prabirshrestha:
@mikoto2000 Can you confirm if this solves your issue.

mikoto2000:
> Can you confirm if this solves your issue.

Yes. This fix solved my issue.

But I got this message after complete.
image

I think needs `silent` add to `execute`.
https://github.com/prabirshrestha/vim-lsp/blob/053f58043b7859717c59134c17b2a4d7ae818e8f/autoload/lsp/utils/text_edit.vim#L63

thomasfaingnaert:
@mikoto2000 Fixed.

prabirshrestha:
Merged. Thanks.

PR #391, opened 2019-05-23
Filter completions

### LINKED ISSUE #377: Got invalid completions with typescript-language-server
Hello!

I've started using vim-lsp for typescript with configuration like this
```vim
if executable('typescript-language-server')
        \ 'name': 'typescript-language-server',
        \ 'cmd': {server_info->[&shell, &shellcmdflag, 'typescript-language-server --stdio']},
        \ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), 'tsconfig.json'))},
        \ 'whitelist': ['javascript', 'javascript.jsx', 'typescript', 'typescript.jsx'],
        \ })
    autocmd FileType typescript,typescript.jsx,javascript,javascript.jsx setlocal omnifunc=lsp#complete
    autocmd FileType typescript,typescript.jsx,javascript,javascript.jsx nnoremap <buffer> <silent> <c-]> :LspDefinition<cr>
    autocmd FileType typescript,typescript.jsx,javascript,javascript.jsx nnoremap <buffer> <silent> <leader>r :LspReferences<cr>
    autocmd FileType typescript,typescript.jsx,javascript,javascript.jsx nnoremap <buffer> <silent> K :LspHover<cr>
endif
```

But, when I triggering completion for partialy typed word I'm getting unexpected results like on the screenshot screenshot below. Completion triggered manually after class App extends Com.
image

```bash
$ vim --version
VIM - Vi IMproved 8.1 (2018 May 18, compiled Mar 21 2019 17:50:56)
Included patches: 1-1027
Compiled by void-buildslave@a-hel-fi
Huge version without GUI.  Features included (+) or not (-):
+acl               +extra_search      +mouse_netterm     +tag_old_static
+arabic            -farsi             +mouse_sgr         -tag_any_white
+autocmd           +file_in_path      -mouse_sysmouse    -tcl
+autochdir         +find_in_path      +mouse_urxvt       +termguicolors
-autoservername    +float             +mouse_xterm       +terminal
-balloon_eval      +folding           +multi_byte        +terminfo
+balloon_eval_term -footer            +multi_lang        +termresponse
-browse            +fork()            -mzscheme          +textobjects
++builtin_terms    +gettext           +netbeans_intg     +textprop
+byte_offset       -hangul_input      +num64             +timers
+channel           +iconv             +packages          +title
+cindent           +insert_expand     +path_extra        -toolbar
+clientserver      +job               +perl              +user_commands
+clipboard         +jumplist          +persistent_undo   +vartabs
+cmdline_compl     +keymap            +postscript        +vertsplit
+cmdline_hist      +lambda            +printer           +virtualedit
+cmdline_info      +langmap           +profile           +visual
+comments          +libcall           -python            +visualextra
+conceal           +linebreak         +python3           +viminfo
+cryptv            +lispindent        +quickfix          +vreplace
+cscope            +listcmds          +reltime           +wildignore
+cursorbind        +localmap          +rightleft         +wildmenu
+cursorshape       +lua               +ruby              +windows
+dialog_con        +menu              +scrollbind        +writebackup
+diff              +mksession         +signs             +X11
+digraphs          +modify_fname      +smartindent       +xfontset
-dnd               +mouse             +startuptime       -xim
-ebcdic            -mouseshape        +statusline        -xpm
+emacs_tags        +mouse_dec         -sun_workshop      +xsmp_interact
+eval              -mouse_gpm         +syntax            +xterm_clipboard
+ex_extra          -mouse_jsbterm     +tag_binary        -xterm_save
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
      user exrc file: "$HOME/.exrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
  fall-back for $VIM: "/usr/share/vim"
Compilation: cc -c -I. -Iproto -DHAVE_CONFIG_H     -fstack-clash-protection -mtune=generic -O2 -pipe   -g -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1
Linking: cc   -L. -Wl,-z,relro -Wl,-z,now     -fstack-protector-strong -rdynamic -Wl,-export-dynamic -Wl,-E  -Wl,-z,relro -Wl,-z,now     -L/usr/local/lib -Wl,--as-needed -o vim    -lSM -lICE -lXt -lX11 -lXdmcp -lSM -lICE  -lm -lelf -lnsl   -lncursesw -lacl -lattr -ldl  -L/usr/lib -llua -Wl,-E -Wl,-z,relro -Wl,-z,now -Wl,--as-needed     -pthread  -L/usr/lib/perl5/core_perl/CORE -lperl -lm -lcrypt -lpthread -ldl  -L/usr/lib/python3.6/config-3.6m-x86_64-linux-gnu -lpython3.6m -lpthread -ldl -lutil -lm  -lruby -lm
```

My vim configuration https://github.com/sivakov512/.vim/blob/master/vimrc#L106

### LINKED ISSUE #438: Use filterText for completion filtering
According to the LSP specification, if a server response to a completion request contains a `filterText` field, this should be used for filtering instead of the `label`. (E.g., if completion is triggered on `foo` and the server returns a completion item with
```
{
    "label": "bar",
    "filterText": "foo bar",
}
```
the list of completions should contain `bar`.

thomasfaingnaert:
This PR filters completion items: items that do not start with the already typed text are discarded.
Here, "already typed text" means the characters in `'iskeyword'` in front of the cursor.

Fixes #377.
Fixes #438.


autoload/lsp/omni.vim
```diff
@@ -140,11 +140,23 @@ function! s:get_completion_result(data) abort
         let l:incomplete = 0
     endif
 
+    let l:current_line = strpart(getline('.'), 0, col('.') - 1)
+    let l:start_pos = match(l:current_line, '\k\+$')
```
prabirshrestha: we might need to make the regex configurable in the future but for now is ok. When someone complains we can make it configure per filetype.

prabirshrestha:
@thomasfaingnaert  could you also fix merge conflicts here.

thomasfaingnaert:
@prabirshrestha The conflicts are due to #392, which does something very similar to `s:filter_match` of this PR. Would changing this line:
https://github.com/prabirshrestha/vim-lsp/blob/3b7ae8c2c618e97df4fcf752f5fb5c9aa879251d/autoload/lsp/omni.vim#L85
to
```vim
return l:label =~ l:match_pattern 
```
(which uses `'ignorecase'`) and then using an `ours` merge strategy be OK for you?

clason:
Just a note: with this PR, I get the following error
```
Error detected while processing function lsp#complete[1]..lsp#omni#complete:
line   32:
E121: Undefined variable: v:none
```
(twice), it also doesn't show candidates that should match (but they all contain colons; that may be a separate issue).

mattn:
AFAIK, neovim does not have v:none.

prabirshrestha:
Might be return empty string?

thomasfaingnaert:
Fixed Neovim problems and merge conflicts.

thomasfaingnaert:
@clason 
That would be broken, yes. It would probably be better to disable this by default, and add a `filter` key (or something similar) in the server config, with possible values `none` (default), `prefix` (this PR), `fuzzy` (for later extensions), etc.

clason:
Yes, that is a good idea -- for example, ncm2 does something similar: https://github.com/ncm2/ncm2/blob/e5a7976ad175251a96c537488d2d9557fafdcc8b/doc/ncm2.txt#L128

thomasfaingnaert:
Added two configuration options:
- `typed_pattern`: used to customise the `'\k*$'` regex.
```vim
" ...
'config': { 'typed_pattern': '\k*$' }
" ...
```
- `filter`: used to specify how to filter completion items. Possible values are `none` or `prefix`. These are specified in a Vim dict to allow later extensions with parameters.
```vim
" ...
'config': { 'filter': { 'name': 'prefix' } }
" ...
```

mattn:
Could you please fix conflict?

thomasfaingnaert:
@mattn Fixed.

mattn:
Sorry, once more please.

thomasfaingnaert:
@matn Done.

sivakov512:
Will it be merged soon?

autoload/lsp/omni.vim
```diff
@@ -63,26 +64,65 @@ function! lsp#omni#complete(findstart, base) abort
...
+        return trim(a:item['word'])
+    endif
+
+    let l:user_data = json_decode(a:item['user_data'])
+    return trim(get(l:user_data, s:user_data_filtertext_key, a:item['word']))
+endfunction
+
+function! s:prefix_filter(item, last_typed_word) abort
+    let l:label = s:get_filter_label(a:item)
+    let l:match_pattern = '^' . a:last_typed_word
```
mattn: please use `stridx(l:label, a:last_typed_word) == 0` or  `stridx(tolower(l:label), tolower(a:last_typed_word)) == 0` instead.

thomasfaingnaert: @mattn Changed. Is this ready to be merged?

clason:
@mattn @thomasfaingnaert There seems to be an issue with `filterText` containing periods: if I enter, e.g., `2.1` (a valid filter text), I see the completion starting after `2.`, and only the `1` seems to be used for filtering (returning too many candidates).

clason:
I assume I need to adapt the `typed_pattern`, but the one I used for `ncm2` and `vim-tex` autocomplete (where this works) doesn't seem to do the trick.

thomasfaingnaert:
@clason What pattern is used in `ncm2` and `vim-tex`? If it matches a `.`, it should work. It might have to do with regex magic setting, resulting in `.` to be interpreted as any char, rather than literal `.`.

clason:
@thomasfaingnaert it's `\w+`. Regex magic is dark magic...

thomasfaingnaert:
@clason `\w` shouldn't match `.`. You could add `.` explicitly by using something like `\(\w\|\.\)\+` (ugly, I know).

clason:
The filtering might actually be done somewhere else; let me check in the `vimtex` sources.

thomasfaingnaert:
That looks a little weird, what is your aim with that regex?

clason:
It is weird. It's not mine; I'm trying to find out how `vimtex` matches `2.1`. It's complicated...

clason:
Ugly or not, your suggestion is working. However, I notice another issue: the `contains` filter doesn't seem to match initial patterns: `lem` doesn't match, e.g., `lem:existence`, but matches `eq:problem`.

clason:
I also have to include the `:` explicitly, even though it's included in `iskeyword`?

thomasfaingnaert:
`\k` uses `iskeyword`, while `\w` just matches alphanumeric chars (at least on Vim).
See `:h /\k` and `:h /\w` for more info, or `:h regexp` if you're bored :smile:

clason:
Yep, just saw that the regex is no longer based on `k`. 🤦‍♂

thomasfaingnaert:
What do you mean with trigger character, exactly?

I think I know why `lem` doesn't match `lem:existence`, will push fix in a minute.

clason:
The characters which tell the server that it should be looking for completion items; servers announce these in their `capabilities` response in, e.g.,
```json
"completionProvider": {
    "triggerCharacters": ["\\", "{", "}", "@", "/", " "]}
}
```

thomasfaingnaert:
You can manually place them in a character class, I guess: `[\{}@/ ].*$`, but you need to make the matching non-greedy. I'll experiment a bit and let you know.

thomasfaingnaert:
How about `[\/{}@/ ][^\{}@/ ]\{-}$`?
Breakdown:
- `[\/{}@/ ]`: the trigger char
- `[^\{}@/ ]\{-}`: any character except for a possible trigger char, repeated 0 or more times, matched non-greedily
- `$`: current cursor position

clason:
I was actually thinking whether it wouldn't make sense for `vim-lsp` to use that as the default (if provided by the server). I think that's how `ncm2-vim-lsp` does it.

thomasfaingnaert:
Hmm, that might be possible. Not sure how we should handle the case where the trigger char is not found. I guess if `<SPACE>` is in the trigger chars, it shouldn't matter.

clason:
Don't I need to escape these chars? (Just putting that regex in `typed_pattern` doesn't work.)

thomasfaingnaert:
It shouldn't need escaping if you use single-quoted rather than double-quoted strings.

clason:
I'm using 
```vimL
if executable('texlab')
                \ 'name': 'texlab',
                \ 'cmd': {server_info->['texlab']},
                \ 'config': {'typed_pattern': '[\/{}@/ ][^\{}@/ ]\{-}$', 'filter': {'name': 'contains'}},
                \ 'whitelist': ['bib','tex'],
                \ })
endif
```

thomasfaingnaert:
How about this: if the `typed_pattern` is empty (which will become the default then, I suppose?), use the trigger char rule, and otherwise use the given pattern?

clason:
Yes, that was what I had in mind. An override is always useful.

thomasfaingnaert:
Ah, for the pattern: completely forgot to exclude the trigger char itself from the filtering: you probably want `[\/{}@/ ]\zs[^\{}@/ ]\{-}$`.

mattn:
@prabirshrestha We can implement `triggerCharacters` with InsertCharPre like signature help. thoughts?

clason:
@thomasfaingnaert Yes, that seems to work in all cases I've mentioned so far!

thomasfaingnaert:
@clason I've updated the PR. If you remove the `typed_pattern` (or set it to the empty string), `vim-lsp` should use the trigger chars by default.

clason:
Indeed, works like a charm!

thomasfaingnaert:
Updated the documentation.

@mattn Anything else that should be changed before merging?

clason:
@thomasfaingnaert If we want to put further customizations in the `config`, would it make sense to scope this with a `completion`?

thomasfaingnaert:
@clason We could, but that might make the server registration unnecessarily verbose and nested. 
@prabirshrestha What do you think?

thomasfaingnaert:
@mattn / @prabirshrestha Is this ready to be merged?

thomasfaingnaert:
Actually, I just upgrade clangd, and this seems to break completion there, so we should hold off on merging for now.

thomasfaingnaert:
@clason Using the trigger characters gives a bit of a problem in `clangd`, since it doesn't define any.
As a result, anytime you start completion, the entire line is cleared:
Peek 2019-09-11 13-29

I guess we should revert to `\k*$` by default?

clason:
OK, that is a problem. I think it would be better to test for the server response and if it is empty, default to  `\k*$` (i.e., `config` > `triggerCharacters` > `\k*$`)?

autoload/lsp/capabilities.vim
```diff
@@ -135,3 +135,15 @@ function! lsp#capabilities#get_signature_help_trigger_characters(server_name) ab
     endif
     return []
 endfunction
+
+function! lsp#capabilities#get_completion_trigger_characters(server_name) abort
+    let l:capabilities = lsp#get_server_capabilities(a:server_name)
+    if !empty(l:capabilities) && has_key(l:capabilities, 'completionProvider')
+        if type(l:capabilities['completionProvider']) == type({})
+            if has_key(l:capabilities['completionProvider'], 'triggerCharacters')
+                return l:capabilities['completionProvider']['triggerCharacters']
```
clason: basically, check here if the return is missing or empty, and then set `\k*$`?

thomasfaingnaert:
The thing is that something similar may still happen even if the server has triggerCharacters.
For example, I can imagine that operators would not be in triggerChars, so if you type
```
f(a, 1 + 2 + <invoke completion here>)
```
then you would delete the entire typed expression `1 + 2 +`.

clason:
But isn't space usually a trigger character? Then this shouldn't happen, right?

thomasfaingnaert:
`f(a,1+2+<completion>)` would still break, though.
My point is, there are probably too many edge-cases to make this default?

clason:
The question for me is whether this is an issue with the LSP specification or with the vim-lsp implementation -- in the first case, I think it makes sense to still follow it since that is what is expected.

Otherwise I'd propose putting this behind a server option `completion: {'use_trigger_chars': v:true}`. My vote is always for configurability over automagic "do what you think I mean" :wink:

autoload/lsp/capabilities.vim
clason: In any case, catching empty trigger chars in `omni#complete()` is probably a good idea...

thomasfaingnaert:
Well, trigger characters are not meant to be used this way at all: their purpose is just to have point where to start completion automatically, but the user can still invoke completion manually (in VSCode with CTRL-SPACE). Therefore, (ab)using it to determine the last lexical token that was typed is not ideal. Not sure how to solve your issue with `1.2` then, though. :(

clason:
Hmm. That's a valid point. Even with that (ab)use, I'd run into problems with `\cite{paper,bo<c-x c-o>`.

Not sure how to solve this. Isn't there a way to get the complete pattern from the server?

thomasfaingnaert:
I don't think there is. Do you have an idea how VSCode does this?

clason:
No, just how ncm2 does it (https://github.com/ncm2/ncm2-vim-lsp/blob/5feab9e3334d5a6a427b19c0ce39c7689d5b6a4d/plugin/ncm2_vim_lsp.vim#L42).

Maybe you could temporarily add the trigger chars (if they exist) to `iskeyword`?

thomasfaingnaert:
I wouldn't touch iskeyword in vim-lsp, as it's a setting that's used quite a lot.
I don't think VSCode does the magic of figuring out what "the user has typed", and placing the cursor at the start. It seems to send completion requests with the complete buffer content and on the exact position of the cursor.

Not sure why #377 doesn't appear in VSCode, though. I should probably look into how VSCode does it first?

thomasfaingnaert:
Back to the drawing board, I guess.

clason:
> I wouldn't touch iskeyword in vim-lsp, as it's a setting that's used quite a lot.

Maybe; I thought the time frame would be short enough, but better safe than sorry.

>  I should probably look into how VSCode does it first?

Or whether that's the job of the extension.

thomasfaingnaert:
> Or whether that's the job of the extension.

Yeah, it appears VSCode doesn't use the vanilla TS language server. You don't have experience with another language client, perchance? :)

clason:
No, there's only one for me :) (The alternative would be natebosch/vim-lsc, but I don't think they have implemented `textEdit`.) I guess your best bet would be autozimu/LanguageClient-neovim.

clason:
@thomasfaingnaert What about manually storing a `startcol` when the completion request is first triggered? Would that work?

thomasfaingnaert:
@clason That looks scary :) One thing I was thinking about is using the start position of the main textedit of a completion item. If that's not supported, we can fallback to a user-configurable `typed_pattern`, defaulting to `\k*$` I suppose.

We would have to think about what we do if different completion items have different starting ranges. I guess we can take the minimum and then "patch" the other completion items. E.g. if I have typed "abcx" and the language server returns "xyz, starting at position 3" and "foo, starting at position 0", we would show the completion menu at position 0 (the minimum), and alter "xyz, starting at position 3" to "abcxyz, starting at position 0". (In practice, I don't think this will happen; so this may be overkill).

Anyway, once we have the position the completion starts at, we can filter using the substring between that position and the cursor.

Not sure how this will interact with asyncomplete, though.
I also have the feeling I'm overthinking this.

thomasfaingnaert:
If we use the column when a completion request is first triggered, we have the same problem, because the user can do it manually. For `xy<C-x><C-o>` we would return the position after `y`, even though we would like the position before `x`.

thomasfaingnaert:
VSCode seems to do something similar (https://github.com/microsoft/vscode/blob/9089a79ecf76a25fccc7c232d80cdd374a7cc66e/src/vs/editor/contrib/suggest/completionModel.ts#L160-L168), haven't looked into it in too much detail, though.

clason:
> @clason That looks scary :)

Black magic, didn't I say ;)

> One thing I was thinking about is using the start position of the main textedit of a completion item. If that's not supported, we can fallback to a user-configurable `typed_pattern`, defaulting to `\k*$` I suppose.

Yes, that could work.

> We would have to think about what we do if different completion items have different starting ranges. I guess we can take the minimum and then "patch" the other completion items. E.g. if I have typed "abcx" and the language server returns "xyz, starting at position 3" and "foo, starting at position 0", we would show the completion menu at position 0 (the minimum), and alter "xyz, starting at position 3" to "abcxyz, starting at position 0". (In practice, I don't think this will happen; so this may be overkill).

Maybe it's possible to crib from this? https://github.com/ncm2/ncm2/commit/b2cf98e4a633296fda4a418255c383e937bb4622

> Anyway, once we have the position the completion starts at, we can filter using the substring between that position and the cursor.

Exactly.

> Not sure how this will interact with asyncomplete, though.
> I also have the feeling I'm overthinking this.

Yeah, it's weird that it Just Works:tm: with `ncm2-vim-lsp`.

thomasfaingnaert:
> Yes, that could work. At least with texlab, the `textEdit` range is always the complete text that is used for matching, so one could just remove that and insert the full completion item.

I've redone the filtering a bit: now it uses the textEdit start position, or if that's not available 
 
> Maybe it's possible to crib from this? [ncm2/ncm2@b2cf98e](https://github.com/ncm2/ncm2/commit/b2cf98e4a633296fda4a418255c383e937bb4622) Not sure it's actually being used, though.

I don't think you can do that with Vim's omnifunc: you have to specify the completion start position when you fill the completion popup; you can't change it later.


There's another issue I noticed: while the filtering applies correctly immediately after you press `<C-x><C-o>`, once you start typing, Vim filtering will take over, which is inherently prefix based:

Peek 2019-09-11 20-56


I think we should probably only support `prefix` filtering, and leave the other filtering options to an external completion framework (be it ncm2, asyncomplete, or others)?

clason:
@thomasfaingnaert Nice! Works like a charm in `texlab`, in all situations I've brought up so far.

I see what you mean: I can't narrow matches by substring; completion stops and I have to retrigger `<c-x c-o>`. I still think allowing substring matches is useful if only for the first triggering and I would hate to lose it; if this restriction was documented, I think it's OK to leave it as is. After all, this is exactly the same behavior as with other omnicomplete functions (such as the one provided by vimtex).

thomasfaingnaert:
@clason 
Added to the doc. Anything you think I should change, or is this ready?
(Apart from changing the failing tests, that is)

clason:
@thomasfaingnaert Actually, you can continue to filter -- as you say -- as long as you stick to prefix. So this note should rather be specific to the `contains` filter. Maybe the following?

`Note that after triggering completion, further filtering is only possible by adding to the prefix (even with the `contain` filter).`

Otherwise looks good to me! Of course, it would be good to test with more servers...

thomasfaingnaert:
@clason That's indeed clearer, I'll change it. I've tested this with clangd 7, clangd 8, typescript. I'll try pyls and eclipse LS next.

clason:
I gave pyls (the only other server I have) a brief run; it worked but I couldn't get `contains` matching?

thomasfaingnaert:
@clason Might be that `pyls` already inherently does a prefix filtering.

thomasfaingnaert:
Seems like it:
```python
def foobarbaz(x, y, z):
    pass

def bar():
    pass

ba<C-x><C-o>
```

returns
```
Wed 11 Sep 2019 21:58:47 CEST:["<---",2,"pyls",{"response":{"id":2,"jsonrpc":"2.0","result":{"isIncomplete":false,"items":[{"label":"bar()","insertText":"bar","sortText":"abar","kind":3,"documentation":"bar()","detail":"test"},{"label":"BaseException","insertText":"BaseException","sortText":"aBaseException","kind":7,"documentation":"BaseException(\\*args: object)\n\nCommon base class for all exceptions","detail":"builtins"}]}},"request":{"id":2,"jsonrpc":"2.0","method":"textDocument/completion","params":{"textDocument":{"uri":"file:///home/thomas/test.py"},"position":{"character":2,"line":6}}}}]
```

clason:
Yes, that sounds reasonable. One of the more frustrating aspects of LSP: not being sure whether something is supposed to be server or client side, so you end up with inconsistencies...

prabirshrestha:
@thomasfaingnaert could you fix the merge conflicts?

thomasfaingnaert:
@prabirshrestha Fixed. Quickly tested this with typescript, python and C++, and it seems to still work.
By the way, I've noticed that the Windows CI recently stopped working. Have you looked into that yet?

prabirshrestha:
Does it also work with asyncomplete when autopopup is set?

I have noticed windows build failing but haven’t looked much into it yet. I will have a look at it by end of this month.

thomasfaingnaert:
@prabirshrestha Seems to work fine with asyncomplete:
Peek 2019-11-05 10-41

mattn:
I'm okay to merge this.

prabirshrestha:
Merged. Thanks.

Sorry for getting to long to get this in. Seems like the PR started on March.

clason:
@prabirshrestha since you asked above: I think only `prefix` filtering works with asyncomplete, not `contains`. I couldn't immediately see from the code why that should be the case, though -- do you have an idea?

thomasfaingnaert:
Maybe asyncomplete does prefix-like filtering internally?

clason:
You're right, there's a default preprocessor that seems to do some filtering: https://github.com/prabirshrestha/asyncomplete.vim/blob/db3ab51ef6d42ac410afaea53fc0513afd0d5e25/autoload/asyncomplete.vim#L445

You can override it, but only globally, I think.

PR #408, opened 2019-06-17
fix `No matching autocommands` message on `doautocmd User lsp_complete_done`

### LINKED ISSUE #407: Too many `No matching autocommands` messages
After https://github.com/prabirshrestha/vim-lsp/commit/20f5b7541a29dd57cebf9c1c1031f8d4320ca275, `No matching autocommands` message is displayed every time I stop the cursor (precisely, after completion is done).
Here's a quote from `:h User`.

> Note that when `:doautocmd User MyEvent` is
> used while there are no matching autocommands,
> you will get an error.  If you don't want
> that, define a dummy autocommand yourself.

itchyny:
This fixes #407.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -25,6 +25,7 @@ augroup _lsp_silent_
     autocmd User lsp_unregister_server silent
     autocmd User lsp_server_init silent
     autocmd User lsp_server_exit silent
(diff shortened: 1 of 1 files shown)


mattn:
Thanks for fixing this.

PR #412, opened 2019-06-25
Add LspPeekDefinition

### LINKED ISSUE #411: Add "peek definition"
I'd be nice to add a `LspPeekDefinition` (insert bike-shedding here, name taken from VS Code that offers this) that works like `LspDefinition`, but instead of jumping to the location in a new buffer, only shows the location in a preview window (like `LspHover` uses; bonus points for floating window when that is implemented).

thomasfaingnaert:
Adds `:LspPeekDefinition`, which displays the definition in the preview window, similar to VSCode's peek definition.

Fixes #411

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -177,7 +177,7 @@ function! s:open_preview(data) abort
     return l:winid
 endfunction
 
-function! lsp#ui#vim#output#preview(data) abort
+function! lsp#ui#vim#output#preview(data, options) abort
     if s:winid && type(s:preview_data) == type(a:data)
        \ && s:preview_data == a:data
        \ && type(g:lsp_preview_doubletap) == 3
@@ -196,11 +196,40 @@ function! lsp#ui#vim#output#preview(data) abort
     let s:preview_data = a:data
     let l:lines = []
     let l:ft = s:append(a:data, l:lines)
+
+    if has_key(a:options, 'filetype')
+        let l:ft = a:options['filetype']
+    endif
+
     call s:setcontent(l:lines, l:ft)
 
     " Get size information while still having the buffer active
     let l:bufferlines = line('$')
     let l:maxwidth = max(map(getline(1, '$'), 'strdisplaywidth(v:val)'))
+
+    " Don't use 'scrolloff', it might mess up the cursor's position
+    let &l:scrolloff = 0
+
+    " Set options of preview window
+    if has_key(a:options, 'statusline')
+        let &l:statusline = a:options['statusline']
+    endif
+
+    if has_key(a:options, 'cursor')
+        execute printf('call cursor(%d, %d)', a:options['cursor']['line'], a:options['cursor']['col'])
+
+        if has_key(a:options['cursor'], 'align')
+            let l:align = a:options['cursor']['align']
+
+            if l:align == "top"
+                normal! zt
+            elseif l:align == "center"
+                normal! zz
+            elseif l:align == "bottom"
+                normal! zb
+            endif
+        endif
+    endif
 
     if g:lsp_preview_keep_focus
       " restore focus to the previous window
@@ -227,7 +256,7 @@ function! s:append(data, lines) abort
 
         return 'markdown'
     elseif type(a:data) == type('')
-        call extend(a:lines, split(a:data, "\n"))
+        call extend(a:lines, split(a:data, "\n", v:true))
 
         return 'markdown'
     elseif type(a:data) == type({}) && has_key(a:data, 'language')
@@ -237,7 +266,7 @@ function! s:append(data, lines) abort
 
         return 'markdown'
     elseif type(a:data) == type({}) && has_key(a:data, 'kind')
-        call extend(a:lines, split(a:data.value, '\n'))
+        call extend(a:lines, split(a:data.value, '\n', v:true))
 
         return a:data.kind ==? 'plaintext' ? 'text' : a:data.kind
     endif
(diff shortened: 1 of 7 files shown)


prabirshrestha:
Can you make use of the changes that came in this PR which just got merged. https://github.com/prabirshrestha/vim-lsp/pull/395

thomasfaingnaert:
I pushed a basic version that works using:
- preview:
preview

- Vim popups:
vim popup

- Neovim floats:
neovim float

prabirshrestha:
@jerdna-regeiz could you have a look at this PR first.

ftplugin/lsp-hover.vim
```diff
@@ -7,13 +7,13 @@ else
   setlocal previewwindow buftype=nofile bufhidden=wipe noswapfile nobuflisted
 endif
 setlocal nocursorline nofoldenable
+setlocal nonumber norelativenumber
```
jerdna-regeiz: why not just put it in the line above?
```suggestion
setlocal nocursorline nofoldenable nonumber norelativenumber
```

plugin/lsp.vim
```diff
@@ -63,7 +64,8 @@ command! LspPreviousReference call lsp#ui#vim#references#jump(-1)
 
 nnoremap <plug>(lsp-code-action) :<c-u>call lsp#ui#vim#code_action()<cr>
 nnoremap <plug>(lsp-declaration) :<c-u>call lsp#ui#vim#declaration()<cr>
-nnoremap <plug>(lsp-definition) :<c-u>call lsp#ui#vim#definition()<cr>
+nnoremap <plug>(lsp-definition) :<c-u>call lsp#ui#vim#definition(0)<cr>
+nnoremap <plug>(lsp-peek-definition) :<c-u>call lsp#ui#vim#definition(1)<cr>
```
jerdna-regeiz: Could we use `v:true` and `v:false` to express more clearly that it is boolean? Or does it really require an int somewhere I missed? (same goes for the command)

jerdna-regeiz:
When `win_execute` is implemented in nvim we should use that instead of switching focus multiple times (maybe we could even manage without switching at all). But for now I have found no other solution as well :( Variables could be set with `nvim_win_set_{var,option}`, but I have no alternative for calling `zz` and the like for adjustment.

plugin/lsp.vim
prabirshrestha: Old version of neovim doesn’t support v:true and it is buggy too.

jerdna-regeiz: okay, didn't know that

ftplugin/lsp-hover.vim
thomasfaingnaert: That's indeed better

autoload/lsp/ui/vim/output.vim
```diff
@@ -213,9 +272,28 @@ function! lsp#ui#vim#output#preview(data) abort
       if has('nvim')
         call s:adjust_float_placement(l:bufferlines, l:maxwidth)
         call s:add_float_closing_hooks()
+
+        if has_key(a:options, 'cursor')
+          let s:suppress_close = v:true
```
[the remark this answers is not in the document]
thomasfaingnaert: I completely missed that, thanks!

thomasfaingnaert:
@jerdna-regeiz Thanks for reviewing my PR! Regarding my TODO, I tried `win_execute` with `normal! zz` and friends, and that doesn't seem to work. Do you know a way to do the alignment in Vim? At the moment, only setting the `firstline` seems to work, but that's kind of annoying to use because you would have to know the window height in advance, and calculate the position based on that. (At least for center and bottom)

jerdna-regeiz:
@thomasfaingnaert Have you tried to set the option afterwards with `popup_setoptions` like when the resize is done for nvim, because the size/content is known then? 
We should up the version check to patch 8.1.1561 for popup_setoptions to work.

thomasfaingnaert:
There's also quite a bit of complex conditionals, which I'm not a big fan of. Perhaps we should set 3 boolean script variables `s:use_preview`, `s:use_vim_popup`, `s:use_neovim_float` instead of always having to type things like 
```vim
if s:supports_floating && g:lsp_preview_float && !has('nvim')
```

jerdna-regeiz:
> There's also quite a bit of complex conditionals, which I'm not a big fan of. Perhaps we should set 3 boolean script variables `s:use_preview`, `s:use_vim_popup`, `s:use_neovim_float` instead of always having to type things like

I agree, but it could be done in a separate PR to not mix a new feature with refactoring.

thomasfaingnaert:
Tried fixing Vim alignment of the popup window in ccaf0f95b4cc9d9835b272a16c5ca4e27d7e0cf2, but I'm facing one more issue: `l:height` is 1, as if the popup still contains `...`, even though the content should already be set by then, right?

thomasfaingnaert:
Seems like using a timer and specifying the width and height explicitly works.

thomasfaingnaert:
Added Peek commands for implementation, type definition and declaration. I have not been able to test these, because I don't have a language server that supports these, but they should work the same as LspPeekDefinition.

As for LspReferences, I'm not sure if it would make sense to use the popup window for those, because you have to interact with the quickfix list to go through the list of references?

clason:
I think it's fine as is -- if somebody comes up with a use case for `PeekReferences` (I don't see one either), that could be added in a future PR. LGTM!

prabirshrestha:
@thomasfaingnaert  This has merge conflicts. Can you fix it?

thomasfaingnaert:
@prabirshrestha Fixed merge conflicts.
I'm not sure what changed in the meantime, but Vim's floating window isn't positioned well anymore: it seems to place the preview above the cursor even when there are only 2 lines available above:
image
I fixed this in 18fcb436b1ef3fc68437133bc0628e2e60d4fa52 by forcing the popup above/below the cursor depending on if the cursor is in the first or second half of the screen.
Should I change this, or do we leave it as is?

prabirshrestha:
@thomasfaingnaert Current behavior is good.

I had also sent you an email but haven't got any reply. Can you send me an email. You can find my email at https://github.com/prabirshrestha

PR #428, opened 2019-07-10
Sign priority

smhc:
Allows configuring the priority of placed signs - useful when used along-side "signify" and other plugins that make use of the sign column.


smhc:
I'll clean up the commits and try again.

PR #440, opened 2019-07-22
Fix g:lsp_preview_max_width for Neovim

thomasfaingnaert:
In response to https://github.com/prabirshrestha/vim-lsp/pull/416#issuecomment-513576062.

I decided to try a different approach. Rather than hard wrapping the text using `gq` and `'textwidth'`, I softwrap by setting the maxwidth.
The only problem with this, is that we need to explicitly calculate the height of the floating window (as far as I can tell, anyway; if someone know a way where we can avoid this, let me know).

To accomplish this, I iterate over all lines and sum `ceil(line_width / max_width)` (i.e. the number of "virtual" lines needed to show this line).

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -295,9 +295,10 @@ function! lsp#ui#vim#output#preview(data, options) abort
     call s:setcontent(l:lines, l:ft)
 
     " Get size information while still having the buffer active
+    let l:maxwidth = max(map(getline(1, '$'), 'strdisplaywidth(v:val)'))
     if g:lsp_preview_max_width > 0
       let l:bufferlines = 0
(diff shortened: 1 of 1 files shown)


clason:
Yes, this works nicely, thank you! (Arguably, adjusting the height of the floating window would be the right thing to do even if `max_width` isn't set.)

clason:
One nitpick: `max_width` seems also to be taken as `min_width`, i.e., shorter hover texts are padded to `max_width`. This is not critical, though (and irrelevant for me if the height adjustment is done without `max_width`, too).

thomasfaingnaert:
@clason 
Fixed, thanks.

thomasfaingnaert:
@clason Can I merge this in or does it still not work perfectly?

clason:
Yes, please!

PR #441, opened 2019-07-23
Nicer syntax highlighting for hover

thomasfaingnaert:
This conceals bold, italic and bold italic text in hovers.
Using https://github.com/tpope/vim-markdown/pull/144, this also conceals codeblocks.

Demo with typescript-language-server:

image

I don't like the empty lines above the codeblocks (the concealed start and end delimiters).
As far as I can tell, conceal doesn't work across lines. Maybe we can use a custom foldexpr to hide these empty lines?


clason:
I think the issue is that height/width are calculated before applying the conceal. Maybe there's a way to recompute the size of the concealed text?

thomasfaingnaert:
The same thing happens in a regular buffer, so it's not related to the height of the popup I believe.

image

thomasfaingnaert:
I think this is a limitation of Vim's conceal. What we could do is remove the newline after the start, and before the end marker, and then apply a custom conceal.
Might be the simplest solution?

clason:
True; one would have to manually strip the empty lines before that -- which I'm not sure is possible, since the lines aren't really empty.

I was also thinking of width: Before the patch, you'd get a popup like
```
[*Some bold text*]
```
With the patch, it's
```
[Some bold text  ]
```
(This issue seems related to wrapping with `conceal` on, which is a `wontfix` for vim.)

thomasfaingnaert:
Hmm, I suppose we could use `strdisplaywidth` and then correct it for conceal. E.g. subtract 2 if we see `*foo*` somewhere.
Do you know a language server that has bold/italic/... in its hover?

clason:
Yes, `texlab` :) You can hover on `\cite{X}` to get a markdown formatted citation (e.g., journal name in italics).

thomasfaingnaert:
Now that you mentioned it, it would annoy me to no end though :)

clason:
Once you see it, you can't unsee it :) Sorry!

thomasfaingnaert:
Moved the markers to the same line. I changed the markers to `{{{code.begin.<language>}}}` and `{{{code.end.<language>}}}` just to avoid interference with other syntax.
Added custom syntax rule to conceal ends and apply correct syntax highlighting.

Result:

image

thomasfaingnaert:
Works in Neovim too, provided `let g:markdown_fenced_languages = ['typescript=javascript']` is in your vimrc.
As an aside, does Neovim not include typescript syntax files?

image

clason:
@thomasfaingnaert Not yet, vim just added that file (from HerringtonDarkholme/yats.vim) last month. (Neovim merges the runtime file updates from upstream patch-by-patch.) 

The file was added to vim in v8.1.1735, which contains rather a lot of changes that also touch popup functionality, so it might make sense to open an issue/PR to just merge that file?

thomasfaingnaert:
Fixed size issue.

Vim:
image

Neovim:
image

thomasfaingnaert:
The sizing will probably break the `max_width` setting though :(

clason:
Hmm. I'm not too much concerned about the `max_width` as long as I get wrapping of long lines. (Others might disagree, though.)

Would it be possible to refactor the `max_width` to avoid multiple code paths by introducing a `float_width` that is set (once) to `min(buf_width,max_width)`?

thomasfaingnaert:
Yes, so that's the problem: I have to disable `'wrap'` because otherwise concealed text creates empty lines. But because `'wrap'` is not set, long lines won't be wrapped either.
I think I'd best skip the conceal and check if I can explicitly mark certain regions as being a certain language.

clason:
I see. For me the wrapping is much more important than the conceal, so that would be appreciated.

clason:
I did notice something fun with conceal, though, since you're familiar with TeX: For texlab, `PeekDefinition` on an `\eqref{` opens a float showing the source for the equation for which this is the `\label` -- but since there's `conceal` support for TeX, you get a (limited) unicode preview! :smile:

thomasfaingnaert:
@clason Nice, another feature ticked off my todo list :smile: 

I've completely reimplemented the code region highlighting: instead of adding markers, I pass the line number and language of each code block to the syntax file, and add a rule using `/\%l` (line number match).

The width issue with conceal is still there, but I would leave it as is. Now that code blocks no longer have huge concealed markers, the width should only be off by a couple characters at most.

If necessary, I suppose we could do the same thing for `*foo*` and the like, by stripping the asterisks in `setcontent`, and passing line and column information to the syntax file.

clason:
@thomasfaingnaert Nice! 

Regarding the stripping of asterisks:  I agree, the width issue is really major and can be left as is.

One issue I noticed is that the height of the peek floats seems to be a bit wonky in neovim: If `PeekDefinition` is triggered in the top half of the window, the float is opened under the current line and extends to the bottom of the window. If it's in the bottom half, the float is opened above the current line *but the height is the same as would fit to the bottom* (which, needless to say, may be uselessly small)?

thomasfaingnaert:
`:%s/major/minor/g`? :smile: 
Would you mind creating a separate issue for the problem with Peek?
For the exact viewport, I don't know if the language server sends the range of the definition. If it does, it certainly seems possible.

clason:
`^not` :)

thomasfaingnaert:
Ah, if the issue with Peek is indeed introduced with this PR, there's no need to create a new issue. That would be strange, though, since I'm not changing anything regarding positioning.

If texlab implements `LocationLink`, I might be able to try to implement that in vim-lsp.

clason:
Actually, the situation seems to be the same on master (a bit hard to tell since the paragraphs are messed up, though).

clason:
New issue opened: #444

clason:
Maybe make applying `conceal` an option, though? There might be filetypes (or console fonts...) where this has unintended side effects.

thomasfaingnaert:
That's a good point. Do you think a global or per-server option (or both?) would be best for this?

clason:
Good question! Both? With per-server options overriding the global one (which all default to false?) Kind of like the `whitelist`--`blacklist` options for file types.

plugin/lsp.vim
```diff
@@ -30,6 +30,7 @@ let g:lsp_preview_autoclose = get(g:, 'lsp_preview_autoclose', 1)
 let g:lsp_preview_doubletap = get(g:, 'lsp_preview_doubletap', [function('lsp#ui#vim#output#focuspreview')])
 let g:lsp_peek_alignment = get(g:, 'lsp_peek_alignment', 'center')
 let g:lsp_preview_max_width = get(g:, 'lsp_preview_max_width', -1)
+let g:lsp_hover_conceal = get(g:, 'lsp_hover_conceal', 1)
```
thomasfaingnaert: Do you think it would be better to disable this by default?

miuirussia:
how to make a border, as it was in the first screenshots?

thomasfaingnaert:
@miuirussia I guess you're using Neovim? Borders are only implemented for Vim popups at the moment.

clason:
@thomasfaingnaert 

> If texlab implements LocationLink, I might be able to try to implement that in vim-lsp.

texlab has now implemented it; I've opened a separate issue about it: #474

mattn:
Could you please fix conflict?

thomasfaingnaert:
@mattn Fixed.

prabirshrestha:
Doesn't seem to be working in vim8-1906. Says retrieving signature help... and does nothing when using signature help. LspHover does work.https://github.com/prabirshrestha/vim-lsp/pull/479 also seems to regress in this branch.
I'm trying this on archlinux with typescript-language-server

plugin/lsp.vim
prabirshrestha: We can enable it by default or look at the user fold settings i.e. `nofoldenable`?

thomasfaingnaert:
@prabirshrestha Fixed

prabirshrestha:
@thomasfaingnaert This is great. Thanks. Merged.

PR #443, opened 2019-07-23
Fix hover breaking on missing filetype

### LINKED ISSUE #442: Hover breaks in typescript-language-server
If I `:LspHover` in typescript-language-server, I get:

image

i.e. the highlighting is not applied, and the cursor is positioned in the floating window.

I don't have any issues using clangd, though:

image

It probably only occurs if the hover contains markdown code blocks.

thomasfaingnaert:
Fixes #442.
Apparently, the issue was that the `set filetype` threw an exception because typescript syntax files were not found.

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -156,7 +156,10 @@ function! s:setcontent(lines, ft) abort
     endif
 
     setlocal readonly nomodifiable
-    let &l:filetype = a:ft . '.lsp-hover'
+    try
+        let &l:filetype = a:ft . '.lsp-hover'
+    catch
+    endtry
   endif
 endfunction
 


autoload/lsp/ui/vim/output.vim
```diff
@@ -156,7 +156,10 @@ function! s:setcontent(lines, ft) abort
     endif
 
     setlocal readonly nomodifiable
-    let &l:filetype = a:ft . '.lsp-hover'
+    try
```
natebosch: You could also consider

`silent! let &l:filetype = a:ft . '.lsp-hover'`

thomasfaingnaert: Thanks, I like that much more. Changed in #446.

PR #447, opened 2019-07-25
Re-porting vim-lsc's diff logic (for multibyte character)

hrsh7th:
Thanks @natebosch

I re-porting vim-lsc's diff logic. (and small fixed for test)

I checked this PR in my environment bellow case.

1. create empty `ts` file.
```ts
```

2. insert text that include multibyte character.
```ts
const ä = [].
```

3. Should be display right completion results.

In my environment, it was passed.

But I check other text that includes`𐐀` character. (from count-utf16 branch's test case)
It is not pass.
Sorry, I don't have this reason.

### DIFF
--- a/autoload/lsp/utils/diff.vim
+++ b/autoload/lsp/utils/diff.vim
@@ -7,77 +7,77 @@
 "
 " Finds a single change between the common prefix, and common postfix.
 function! lsp#utils#diff#compute(old, new) abort
-  let [start_line, start_char] = s:FirstDifference(a:old, a:new)
-  let [end_line, end_char] =
-      \ s:LastDifference(a:old[start_line :], a:new[start_line :], start_char)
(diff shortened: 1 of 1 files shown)


clason:
Sorry, that doesn't fix the issue with the wrong offsets in `pyls` -- but at least it doesn't crash anymore, so this is progress!

mattn:
Please add local scope indicator `l:` in function scope.

mattn:
I don't know neovim well, strchars and strgetchar are supported on neovim?

thomasfaingnaert:
@mattn They are, yes.

mattn:
Could you please add test?

mattn:
LGTM

mattn:
LGTM

aktau:
Next time it would be good to squash these commits. Now it looks like the diff is mostly about first removing and then adding `l:`.

PR #449, opened 2019-07-26
Fix tagstack

daisuzu:
The tagstack introduced in #435 has no feature to cleanup items.
For example, the following case will unexpected results.

```vim
:LspDefinition
:LspDefinition
:pop
:pop
:LspDefinition
:pop

" Expect the tagstack to be at the bottom.
" But possible to go back somewhere when next :pop.
```

In addition, set `bufnr` of the item.

### DIFF
--- a/autoload/lsp/ui/vim.vim
+++ b/autoload/lsp/ui/vim.vim
@@ -419,31 +419,31 @@ function! s:handle_symbol(server, last_req_id, type, data) abort
 endfunction
 
 function! s:update_tagstack() abort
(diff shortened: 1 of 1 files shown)


mattn:
Could you please extract small function to do the part?

daisuzu:
@mattn 

Thank you for comment.
I fixed and refactored a bit.

mattn:
LGTM.

mattn:
Thank you

daisuzu:
Thank you for merging! :)

PR #451, opened 2019-07-27
Support utf16 position

hrsh7th:
I tried to support utf16 character position.

Currently, It used to get position of diff or column (Sorry, highlight positioning is not support now).

In my environment, I checked below case and works well.

1. open empty ts file.
```ts
```

2. add character that is 2 code units.
```ts
const obj = { '𐐀': [] as string[] };
obj['𐐀'].#
```
※ `#` is cursor position.

3. should be display right completion results.
- `reduce`, `forEach` etc...

4. And remove all lines.

This use mattn's utf16 code units counting function.

Thanks mattn.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -729,7 +729,7 @@ function! lsp#get_text_document_identifier(...) abort
 endfunction
 
 function! lsp#get_position(...) abort
-    return { 'line': line('.') - 1, 'character': col('.') -1 }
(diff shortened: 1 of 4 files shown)


hrsh7th:
The logic in master branch is not support 2 code units character.
I hope this PR solve this problem.

hrsh7th:
This has bug. Sorry, I close this at the moment.

PR #464, opened 2019-08-11
Implement semantic highlighting

thomasfaingnaert:
This PR implements the unofficial extension to the LSP protocol for semantic highlighting (https://github.com/microsoft/vscode-languageserver-node/pull/367).
I've mainly been testing with Eclipse Java LS.

Before:
image

After:
image

<details>
<summary>Program.java</summary>

```java
public class Program {
    public static void Main(String[] args) {
        System.out.println("Hello World");
        Test("a", 3);
    }

    public static void Test(String s, int x) {
    }
}
```
</details>

<details>
<summary>Language server registration</summary>

```vim
if executable('java') && filereadable(expand('~/lsp/eclipse.jdt.ls/plugins/org.eclipse.equinox.launcher_1.5.300.v20190213-1655.jar'))
                \ 'name': 'eclipse.jdt.ls',
                \ 'cmd': {server_info->[
                \     'java',
                \     '-Declipse.application=org.eclipse.jdt.ls.core.id1',
                \     '-Dosgi.bundles.defaultStartLevel=4',
                \     '-Declipse.product=org.eclipse.jdt.ls.core.product',
                \     '-Dlog.level=ALL',
                \     '-noverify',
                \     '-Dfile.encoding=UTF-8',
                \     '-Xmx1G',
                \     '-jar',
                \     expand('~/lsp/eclipse.jdt.ls/plugins/org.eclipse.equinox.launcher_1.5.300.v20190213-1655.jar'),
                \     '-configuration',
                \     expand('~/lsp/eclipse.jdt.ls/config_linux'),
                \     '-data',
                \     getcwd()
                \ ]},
                \ 'whitelist': ['java'],
                \ 'semantic_highlight': {
                \   'entity.name.type.class.java': 'Identifier',
                \   'entity.name.function.java': 'Label',
                \ }
                \ })
endif
```
</details>


autoload/lsp/ui/vim/semantic.vim
```diff
@@ -0,0 +1,183 @@
...
+endfunction
+
+function! s:add_highlight(server, buf, line, highlights) abort
+    let l:scopes = lsp#ui#vim#semantic#get_scopes(a:server)
+
+    if s:use_vim_textprops
+        " Clear text properties from the previous run
+        for l:scope_idx in range(len(l:scopes))
+            let l:scope = l:scopes[l:scope_idx]
+            call prop_remove({'bufnr': a:buf, 'type': s:get_textprop_name(a:server, l:scope_idx), 'all': v:true}, a:line + 1)
```
prabirshrestha: we need to have similar checks here. if the buf is unloaded we should ignore.

autoload/lsp/ui/vim/semantic.vim
```diff
@@ -0,0 +1,187 @@
...
+endfunction
+
+function! s:tokens_to_hl_info(token) abort
+    let l:ret = []
+    let l:octets = lsp#utils#base64_decode(a:token)
+
+    for l:i in range(0, len(l:octets) - 1, 8)
+        let l:char = s:octets_to_number(l:octets[i : i+3])
+        let l:length = s:octets_to_number(l:octets[i+4 : i+5])
+        let l:scope = s:octets_to_number(l:octets[i+6 : i+7])
```
mattn: This is a point to the code. Just question. Where this octets come from?

thomasfaingnaert: Do you mean why I used these specific offsets?

mattn: Yes, it is.

mattn: I don't understand why octet format is required in this part.

thomasfaingnaert: This is what the spec says, see "tokens" in https://github.com/microsoft/vscode-languageserver-node/blob/5a9b33c23de84c3341e011e79221795a8059375b/protocol/src/protocol.semanticHighlighting.proposed.md.
The tokens contain the character (4 bytes), length (2 bytes), and the scope (2 bytes). This is done for all tokens, and the resulting number is base64 encoded. Not sure why they made this process so complicated, but we'll have to work with it, I guess.

mattn: Ah, now I understand why I couldn't find this from the Language Server Specification. Okay, please wait. I'll try this with some language servers. Thanks.

thomasfaingnaert: So far, I've only found eclipse LS. `ccls` seems to use their own implementation.

prabirshrestha:
How does this work with scoped languages such as html with javascript and css or markdown with some code? Not sure if there are servers that support this. 

Do we need a way to disable this feature?

mattn:
I tried this.

before
image

after
image

Something wrong?

thomasfaingnaert:
@mattn You need to link semantic scopes to Vim highlight groups. See my language server registration for Java as an example. You can check the scopes with `:LspSemanticScopes`.

thomasfaingnaert:
By default, I'm using `Ignore`, which probably hides the text on your machine? (On mine, it does nothing).

thomasfaingnaert:
> How does this work with scoped languages such as html with javascript and css or markdown with some code? Not sure if there are servers that support this.
Does the html language server support nesting css, etc. and semantic highlight? If so, I'll test this. In any case, this is something the language server should handle.
 
> Do we need a way to disable this feature?
I guess that would be useful. Shouldn't be too hard to implement: adding an `&& g:lsp_semantic_enabled` to `lsp#ui#vim#semantic#is_enabled` should work.

prabirshrestha:
Another option is to allow default different.

thomasfaingnaert:
> Another option is to allow default different.

What do you mean?

prabirshrestha:
let g:lsp_default_semantic_highlight_group = “Ignore”

thomasfaingnaert:
@prabirshrestha Ah, I see. Is cdb28d8a4cffd13fb5c40635f23425188a5a13c7 sufficient to disable?

thomasfaingnaert:
If the user doesn't specify semantic scopes, we should just revert to the built-in highlight. I though `Ignore` would accomplish this as a dummy highlight that doesn't change anything, but apparently it functions as a conceal.
Is there a way to define such a highlight group?

thomasfaingnaert:
@prabirshrestha While we're doing the winter cleaning of PRs, any update on this? Apart from the merge conflicts, the only thing that needs to be addressed is the "Ignore" highlight group, which results in semantic scopes that do not have an associated highlight group to just be hidden (in some color schemes at least).

prabirshrestha:
This PR I haven't got to look at it thoroughly yet. Currently looking in smaller PRs so I can get it out of my way.

I will be looking at this tonight about to go out now.

Few questions in the mean time:
* Do we also know if other languages supports this yet. I'm ok for now even if only java support this as it would be a good example for other servers to on board.
* Can we have checks on buffer not being unloaded yet since it is async. I think there are still bugs in master that currently has this issue.  Might be move the check in `handle_semantic`?


Trying to cleanup issues and PRs so that I can work on adding new features to vim lsp. If you have any ideas feel free to share either on perf/new features/testing.

thomasfaingnaert:
>     * Do we also know if other languages supports this yet. I'm ok for now even if only java support this as it would be a good example for other servers to on board.

I only know of eclipse jdt and language servers based on lsp4j. cquery also supports semantic highlighting, but it uses a different API last I checked.

>     * Can we have checks on buffer not being unloaded yet since it is async. I think there are still bugs in master that currently has this issue.  Might be move the check in `handle_semantic`?

Seems there already is some check for that in `s:add_highlight` at https://github.com/prabirshrestha/vim-lsp/pull/464/files#diff-b546a57b22f6d7e4a8af08ccda8384a2R63-R65. Not sure why I put it there, because `handle_semantic` is a much better place for it; I'll move it there.

> Trying to cleanup issues and PRs so that I can work on adding new features to vim lsp. If you have any ideas feel free to share either on perf/new features/testing.

Well, one thing I would really like to see is integration tests with a couple language servers.
This will also enable us to add MWEs for bugs we fix to avoid reintroducing them in the future.

I experimented a bit with this for https://github.com/thomasfaingnaert/vim-lsp-ultisnips, but had to abandon it due to time constraints.
The main issue there was that I couldn't just use `normal!` with vim-themis, because of the asynchronicity.
I did get some promising results with running Vim in a tmux session and using `tmux send-keys`, so that may be a possibility for vim-lsp as well.
For Windows, using `WScript.Shell.SendKeys` from COM should work.
We could also reuse the testing framework that Vim uses?

Of course, this is assuming that regular vim-themis tests won't suffice.
I imagine that will be the case though for things like completion, which are inherently asynchronous.

prabirshrestha:
Test is my biggest issue currently now and that was why yesterday was working on tests.
I'm ok with how vim tests if it also works in neovim and is as easy to use as themis. 

Let us use https://github.com/prabirshrestha/vim-lsp/issues/624 to brainstorm so we don't pollute this. I will add more thoughts on tests in that issue.

prabirshrestha:
@thomasfaingnaert for snippets try this instead. https://github.com/hrsh7th/vim-vsnip and https://github.com/hrsh7th/vim-vsnip-integ. I recently started using it and love it. It is also written in pure vimscript so works well without python. It is aimed for LSP and vscode format unlike other vim plugins so maximum compatibility.

thomasfaingnaert:
@prabirshrestha Thanks, I'll check it out. Does it support all LSP snippet features already?

thomasfaingnaert:
Allright, should be good to go now.

thomasfaingnaert:
CI is failing because `l:i` is unused in:
```vim
for l:i in range(3)
    call add(l:octets, l:group_dec % 256)
    let l:group_dec = l:group_dec / 256
endfor
```

I guess I can rewrite it using a while...

prabirshrestha:
@thomasfaingnaert I think this was brought yesterday too. Is it possible to prefix the variable with underscore so the lint ignores or manually add a comment to that line for it to ignore. else can use while loop.

autoload/lsp/ui/vim/semantic.vim
```diff
@@ -0,0 +1,194 @@
...
+    if lsp#client#is_error(a:data['response'])
+        call lsp#log('Skipping semantic highlight: response is invalid')
+        return
+    endif
+
+    let l:uri = a:data['response']['params']['textDocument']['uri']
+    let l:path = lsp#utils#uri_to_path(l:uri)
+    let l:bufnr = bufnr(l:path)
+
+    if l:bufnr < 0 | return | endif
```
prabirshrestha: can we merge this check with `!bufloaded`? Or should `bufloaded` check be enough?

autoload/lsp/ui/vim/semantic.vim
```diff
@@ -0,0 +1,194 @@
...
+        for l:highlight in a:highlights
+            call nvim_buf_add_highlight(a:buf, s:namespace_id, s:get_hl_name(a:server, l:scopes[l:highlight['scope']]), a:line, l:highlight['char'], l:highlight['char'] + l:highlight['length'])
+        endfor
+    endif
+endfunction
+
+function! s:get_hl_name(server, scope) abort
+    let l:hl = 'LspUnknownScope'
+
+    try
```
prabirshrestha: curious what is the reason for `try..catch` here?

might be good to comment it.

prabirshrestha:
Based on this clangd also supports the same semantic highlight protocol. can we check if it works there too?

https://clang.llvm.org/extra/clangd/Features.html#complete-list-of-features

autoload/lsp/ui/vim/semantic.vim
thomasfaingnaert: Seems `bufloaded` already returns false if `bufnr` is `-1`, so only using `bufloaded` is enough.

autoload/lsp/ui/vim/semantic.vim
thomasfaingnaert: The LSP scopes are organised in a tree-like structure:
image

So if I want to look up the highlight group for e.g. `entity.name.function.java > meta.function-call.java > ... > source.java` I need to use the most specific setting that the user has defined. The cleanest way is to just iterate in the order `entity.name.function.java`, `entity.name.function.java > meta.function-call.java`, ..., and catch the exception if an entry doesn't exist.

I can also remove the try and add checks for `has_key(a:scope, l:i)` and `has_key(l:info, 'semantic_highlight')`, but I think the try is a little less cluttered.

thomasfaingnaert:
@prabirshrestha Ah indeed, I'm stuck on clangd 7, which didn't support semantic highlight. I'll try it with 10

thomasfaingnaert:
Seems to work as well:
```
$ clangd --version
clangd version 9.0.0 (tags/RELEASE_900/final)
```

```vim
" ...
            \ 'semantic_highlight': {
            \   'entity.name.function.cpp': 'Label',
            \   'entity.name.function.method.cpp': 'Todo',
            \   'entity.name.type.class.cpp': 'Identifier',
            \ },
" ...
```

Before:
image


After:
image

prabirshrestha:
Looks good. Merged. Thanks for the PR.

Barricade:
Hello, I can't make it work for clangd 9.0.1. Jump to definition e.t.c. work fine. Output of `:LspSemanticScopes` - `Semantic highlighting not supported for cpp`.

zoumi:
How to enable semantic highligting for #ifdef with clangd?
This is my setting, but it's seem nothing happened. I'm using gvim 8.2 and clangd 12.

```vim
Plug 'prabirshrestha/asyncomplete.vim'
Plug 'prabirshrestha/asyncomplete-lsp.vim'
Plug 'prabirshrestha/vim-lsp'
Plug 'mattn/vim-lsp-settings'
Plug 'prabirshrestha/asyncomplete-ultisnips.vim'
Plug 'prabirshrestha/asyncomplete-tags.vim'
Plug 'prabirshrestha/asyncomplete-buffer.vim'

let g:asyncomplete_auto_completeopt = 0
let g:lsp_documentation_float = 1
let g:lsp_textprop_enabled = 1
let g:lsp_ignorecase = 1
let g:lsp_semantic_enabled = 1
let g:lsp_preview_max_height = 30
let g:lsp_diagnostics_echo_cursor = 1
let g:asyncomplete_matchfuzzy = 0

let g:lsp_settings = {
            \ 'clangd': {
                \ 'semantic_highlight': {'meta.disabled':'Comment'},
                \ },
                \ }
```

PR #470, opened 2019-08-19
Add g:lsp_signs_priority option

clktmr:
When used with other plugins (e.g. gitgutter), the signs will often get
replaced. This patch makes priority configurable for the user and
defaults to a value of 15, which should be a reasonable value for most
use cases.


clktmr:
> Can we use the Vim default of `10` instead?

Sure, sounds more reasonable indeed.

> It might also be worth it to make this configurable per severity level.

I thought about this, but couldn't come up with a sign group that would
fit between the severities.  This would just add unnecessary complexity
in my opinion.

thomasfaingnaert:
> I thought about this, but couldn't come up with a sign group that would fit between the severities. This would just add unnecessary complexity in my opinion.

Priorities are assigned independent of sign group, right? You would just need to make `g:lsp_signs_priority` a dict and use `l:item['severity']` as a key.

But I agree, this may make things unnecessarily complicated. We can always change this later if someone needs it.

thomasfaingnaert:
Merged. Thanks for contributing!

smhc:
I had added support for this and per server/severity mapping under PR https://github.com/prabirshrestha/vim-lsp/pull/429

PR #478, opened 2019-08-27
Update SymbolKind to latest specification

### LINKED ISSUE #477: Update SymbolKinds enums to latest Specification
The `SymbolKinds` enum in autoload/lsp/ui/vim/utils.vim is not up to date (see https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol). 

(Easy fix, but probably better done directly by someone with commit access than via a PR.)

clason:
closes #477

(Fields start with uppercase in specification but in lowercase in the vim-lsp enum; I haven't changed that. Note that two new fields are in camel case: `enumMember` and `typeParameter`; the first looks strange in all lowercase, so I also changed the second for consistency.)

### DIFF
--- a/autoload/lsp/ui/vim/utils.vim
+++ b/autoload/lsp/ui/vim/utils.vim
@@ -62,11 +62,11 @@ let s:symbol_kinds = {
     \ '19': 'object',
     \ '20': 'key',
     \ '21': 'null',    
-    \ '22': 'enumMember',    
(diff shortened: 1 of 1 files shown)


mattn:
I prefer `enum number` instead of `enumMumber`. cc @prabirshrestha

mattn:
Thank you

PR #485, opened 2019-09-04
Announce support for Item and Symbol kinds

### LINKED ISSUE #484: Announce support for completionItemKind
According to the LSP specification, servers will only use the `text` and `references` kinds for completion items if the client doesn't explicitly announce a list of kinds it supports on initialization. Since vim-lsp supports the full range of `kinds` from the current specification, it should announce them.

(This probably goes for `symbolKind` as well.)

clason:
Fixes #484 

Announcing `symbolKind` support while I'm at it.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -380,6 +380,18 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   'workspace': {
     \       'applyEdit': v:true,
     \       'configuration': v:true
+    \   },
+    \   'textDocument': {
+    \       'completion': {
+    \           'completionItemKind': {
+    \              'valueSet': [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25]
+    \           }
+    \       },
+    \       'documentSymbol': {
+    \           'symbolKind': {
+    \              'valueSet': [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]
+    \           }
+    \       }
     \   }
     \ }
 endfunction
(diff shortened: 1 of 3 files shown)


autoload/lsp.vim
```diff
@@ -380,6 +380,18 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   'workspace': {
     \       'applyEdit': v:true,
     \       'configuration': v:true
+    \   },
+    \   'textDocument': {
+    \       'completion': {
+    \           'completionItemKind': {
+    \              'valueSet': [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25]
```
thomasfaingnaert: Can you get the supported kinds from `omni.vim` instead of hardcoding them? That way we won't forget updating this line.

clason: Good idea -- how would I do that in `vimL`?

thomasfaingnaert: You can add a function to `omni.vim` that returns `keys(s:kind_text_mappings)` and call that here.

thomasfaingnaert: Though you may also need to convert the Strings you get to Numbers, as well. `:h map()` will be useful for that.

thomasfaingnaert: You can try `map(keys(...), {idx, key -> str2nr(key)})`.

clason: That works, although the list is sorted wrong (`[10,11,...,1,2,...]`). Seems to make no difference, though.

thomasfaingnaert: `sort(..., {lhs, rhs -> lhs - rhs})` will fix that, though I don't think it matters.

clason: Thanks! I'd suggest leaving it for now -- I'll keep an eye open, and if something weird happens, I now know how to fix it.

thomasfaingnaert:
Merged, thanks for contributing.

clason:
Great, thank you for your help and patience!

PR #497, opened 2019-09-10
Highlight active parameter in signatureHelp

hrsh7th:
Thanks for signatureHelp.

I added to highlights active parameter in signature help.

スクリーンショット 2019-09-10 12 08 32

スクリーンショット 2019-09-10 12 08 46

### DIFF
--- a/autoload/lsp/ui/vim/signature_help.vim
+++ b/autoload/lsp/ui/vim/signature_help.vim
@@ -54,25 +54,35 @@ function! s:handle_signature_help(server, data) abort
             let l:parameters = get(l:signature, 'parameters', [])
             let l:parameter_index = a:data['response']['result']['activeParameter']
             let l:parameter = get(l:parameters, l:parameter_index, {})
-            let l:parameter_label = get(l:parameter, 'label', '')
+            let l:parameter_label = s:get_parameter_label(l:signature, l:parameter)
             if !empty(l:parameter_label)
-                let l:parts = split(l:label, '\V' . escape(l:parameter_label, '\/?'))
-                if len(l:parts) == 2
-                    let l:label = join([l:parts[0], '`', l:parameter_label, '`', l:parts[1]], '')
-                endif
+                let l:label = substitute(l:label, '\V\(' . escape(l:parameter_label, '\/?') . '\)', '`\1`', 'g')
             endif
         endif
 
         let l:contents = [l:label]
+
         if has_key(l:signature, 'documentation')
             call add(l:contents, l:signature['documentation'])
         endif
+
         call lsp#ui#vim#output#preview(a:server, l:contents, {'statusline': ' LSP SignatureHelp'})
         return
     else
         " signature help is used while inserting. So this must be graceful.
         "call lsp#utils#error('No signature help information found')
     endif
+endfunction
+
+function! s:get_parameter_label(signature, parameter) abort
+    if has_key(a:parameter, 'label')
+        if type(a:parameter) == v:t_list
+            let l:string_range = a:parameter['label']
+            return strcharpart(a:signature['label'], l:string_range[0], l:string_range[1])
+        endif
+        return a:parameter['label']
+    endif
+    return ''
 endfunction
 
 function! s:insert_char_pre() abort
--- a/autoload/lsp/ui/vim/signature_help.vim
+++ b/autoload/lsp/ui/vim/signature_help.vim
@@ -62,6 +62,15 @@ function! s:handle_signature_help(server, data) abort
 
         let l:contents = [l:label]
 
+        if exists('l:parameter')
+            let l:parameter_doc = s:get_parameter_doc(l:parameter)
+            if !empty(l:parameter_doc)
+                call add(l:contents, '')
+                call add(l:contents, l:parameter_doc)
+                call add(l:contents, '')
+            endif
+        endif
+
         if has_key(l:signature, 'documentation')
             call add(l:contents, l:signature['documentation'])
         endif
@@ -83,6 +92,19 @@ function! s:get_parameter_label(signature, parameter) abort
         return a:parameter['label']
     endif
     return ''
+endfunction
+
+function! s:get_parameter_doc(parameter) abort
+    if !has_key(a:parameter, 'documentation')
+        return ''
+    endif
+
+    if type(a:parameter['documentation']) == v:t_dict
+        let l:doc = copy(a:parameter['documentation'])
+        let l:doc['value'] = printf('***%s*** - %s', a:parameter['label'], l:doc['value'])
+        return l:doc
+    endif
+    return printf('***%s*** - %s', a:parameter['label'], l:doc)
 endfunction
 
 function! s:insert_char_pre() abort
--- a/autoload/lsp/ui/vim/signature_help.vim
+++ b/autoload/lsp/ui/vim/signature_help.vim
@@ -38,8 +38,32 @@ function! s:handle_signature_help(server, data) abort
     endif
 
     if !empty(a:data['response']['result']) && !empty(a:data['response']['result']['signatures'])
-        let l:signature = a:data['response']['result']['signatures'][0]
-        let l:contents = [l:signature['label']]
+        " Get current signature.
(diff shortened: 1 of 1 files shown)


autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -38,8 +38,32 @@ function! s:handle_signature_help(server, data) abort
...
+
+        " Signature label.
+        let l:label = l:signature['label']
+
+        " Mark current parameter.
+        if has_key(a:data['response']['result'], 'activeParameter')
+            let l:parameters = get(l:signature, 'parameters', [])
+            let l:parameter_index = a:data['response']['result']['activeParameter']
+            let l:parameter = get(l:parameters, l:parameter_index, {})
+            let l:parameter_label = get(l:parameter, 'label', '')
```
thomasfaingnaert: This can also be `[number, number]`, so we should handle that case as well.

autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -38,8 +38,32 @@ function! s:handle_signature_help(server, data) abort
...
+            if !empty(l:parameter_label)
+                let l:parts = split(l:label, '\V' . escape(l:parameter_label, '\/?'))
```
thomasfaingnaert: Can't the split and join be replace by a `substitute()` call?

autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -38,8 +38,32 @@ function! s:handle_signature_help(server, data) abort
...
+                if len(l:parts) == 2
+                    let l:label = join([l:parts[0], '`', l:parameter_label, '`', l:parts[1]], '')
+                endif
+            endif
+        endif
+
+        let l:contents = [l:label]
         if has_key(l:signature, 'documentation')
             call add(l:contents, l:signature['documentation'])
         endif
```
thomasfaingnaert: Can we also add documentation for the current parameter here?

thomasfaingnaert:
As an aside, is that a language server for VimL? I haven't been able to find one, yet.

autoload/lsp/ui/vim/signature_help.vim
thomasfaingnaert: Something like `***<param name>***_: <documentation>` looks good:
image

thomasfaingnaert: Or, if we want to mimic VSCode, `_<documentation>_` above the documentation of the function:
image

autoload/lsp/ui/vim/signature_help.vim
hrsh7th: Thanks for review.
Sorry, I missed `[number, number]`.

I implemented, But I don't know the server that return `[number, number]`.
Could you know the server?

Anyway, I will update this PR.

autoload/lsp/ui/vim/signature_help.vim
hrsh7th: I tried to implement decoration.
What do you think of this?

thomasfaingnaert: Looks great! :smiley:

autoload/lsp/ui/vim/signature_help.vim
thomasfaingnaert: I tested a couple, haven't found one yet, unfortunately. For now, you can test it with a hardcoded value, I guess.

hrsh7th:
Yes, I used `vim-language-server`.

I installed vim-language-server via `npm i -g vim-language-server`.

autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -51,6 +83,30 @@ function! s:handle_signature_help(server, data) abort
...
+endfunction
+
+function! s:get_parameter_doc(parameter) abort
+    if !has_key(a:parameter, 'documentation')
+        return ''
+    endif
+
+    if type(a:parameter['documentation']) == v:t_dict
+        let l:doc = copy(a:parameter['documentation'])
+        let l:doc['value'] = printf('***%s*** - %s', a:parameter['label'], l:doc['value'])
```
hrsh7th: Should I check kind is `markdown`?

thomasfaingnaert: This should work for both cases, I believe, so a check is not necessary.

hrsh7th:
I've fixed for CI and reviews.

autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -51,6 +83,30 @@ function! s:handle_signature_help(server, data) abort
     endif
 endfunction
 
+function! s:get_parameter_label(signature, parameter) abort
+    if has_key(a:parameter, 'label')
+        if type(a:parameter) == v:t_list
+            let l:string_range = a:parameter['label']
+            return strcharpart(a:signature['label'], l:string_range[0], l:string_range[1])
```
thomasfaingnaert: The last argument of `strcharpart` is a length, but LSP returns the exclusive end offset instead.

autoload/lsp/ui/vim/signature_help.vim
```diff
@@ -51,6 +83,33 @@ function! s:handle_signature_help(server, data) abort
...
+
+function! s:get_parameter_doc(parameter) abort
+    if !has_key(a:parameter, 'documentation')
+        return ''
+    endif
+
+    if type(a:parameter['documentation']) == v:t_dict
+        let l:doc = copy(a:parameter['documentation'])
+        let l:doc['value'] = printf('***%s*** - %s', a:parameter['label'], l:doc['value'])
+        return l:doc
```
thomasfaingnaert: Actually, I'm wondering what happens if this is passed to `lsp#ui#vim#output#preview`: no case in `s:append` would apply, so these lines will be skipped. Should we not simply return `printf(...)` here?

autoload/lsp/ui/vim/signature_help.vim
hrsh7th: Thanks!
I had two mistake... and resolved its.

I checked following code.

```VimL
" 012345
echo s:get_parameter_label({ 'label': '012345' }, { 'label': [0, 6] })

" 01234
echo s:get_parameter_label({ 'label': '012345' }, { 'label': [0, 5] })

" 12345
echo s:get_parameter_label({ 'label': '012345' }, { 'label': [1, 6] })

" 1234
echo s:get_parameter_label({ 'label': '012345' }, { 'label': [1, 5] })
```

autoload/lsp/ui/vim/signature_help.vim
hrsh7th: Parameter's documentation is `string or MarkupContent` in LSP spec.

I checked `lsp#ui#vim#output.vim`, it will handle `MarkupContent`.
So I was not modified return format.
(https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/ui/vim/output.vim#L397)

Though if we return string, `output.vim` still handles documentation as markdown.

thomasfaingnaert: Ah nevermind, I missed that `MarkupContent` has a `kind`, so the last case in `s:appends` does indeed apply. Sorry for the confusion!

hrsh7th: No worries.
You code review is very helpful for me! Thanks.

thomasfaingnaert:
@hrsh7th Looks good to me. Anything you'd like to change, or can I merge this in?

hrsh7th:
Yes, I think it's ok to merge.
I have no changes for this PR.

thomasfaingnaert:
@hrsh7th Awesome, thanks for the contribution. Are you planning to tackle #493 next?

hrsh7th:
Yes, I planning to tackle #493.

But I should understand the reason that last my patch was not work.
If I understood the reason, I will tackle.

In fact, in my environment is not work currently signature help.
If remove `l:position.character += 1`, it works well, But I don't know the reason.

So may take some time.

PR #500, opened 2019-09-10
Allow for MarkupContent in completion info

clason:
According to the specification, the language server can provide a `documentation` field for a completion response either as a `string` or as a `MarkupContent`; vim-lsp only supported the former. This change also tests for `MarkupContent` in `plaintext` format (as opposed to `markdown`, which could be an image link) and sets the `info` field accordingly.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -386,6 +386,9 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   },
     \   'textDocument': {
     \       'completion': {
+    \           'completionItem': {
+    \              'documentationFormat': 'plaintext'
+    \           },
     \           'completionItemKind': {
     \              'valueSet': lsp#omni#get_completion_item_kinds()
     \           }
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -387,7 +387,7 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   'textDocument': {
     \       'completion': {
     \           'completionItem': {
-    \              'documentationFormat': 'plaintext'
+    \              'documentationFormat': ['plaintext']
     \           },
     \           'completionItemKind': {
     \              'valueSet': lsp#omni#get_completion_item_kinds()
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -252,7 +252,7 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
             let l:completion['info'] .= a:item['documentation']
         elseif type(a:item['documentation']) == type({}) && 
                     \ has_key(a:item['documentation'], 'value')
-            " field is MarkupContent (string or markdown, we don't care)
+            " field is MarkupContent (hopefully 'plaintext')
             let l:completion['info'] .= a:item['documentation']['value']
         endif
     endif
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -250,7 +250,9 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     if has_key(a:item, 'documentation')
         if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
-        elseif a:item['documentation']['kind'] == 'plaintext' " field is MarkupContent
+        elseif type(a:item['documentation']) == type({}) && 
+                    \ a:item['documentation']['kind'] == 'plaintext'
+            " field is MarkupContent
             let l:completion['info']['value'] .= a:item['documentation']
         endif
     endif
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -253,7 +253,7 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
         elseif type(a:item['documentation']) == type({}) && 
                     \ has_key(a:item['documentation'], 'value')
             " field is MarkupContent (string or markdown, we don't care)
-            let l:completion['info']['value'] .= a:item['documentation']
+            let l:completion['info'] .= a:item['documentation']['value']
         endif
     endif
 
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -248,8 +248,10 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     endif
 
     if has_key(a:item, 'documentation')
-        if type(a:item['documentation']) == type('')
+        if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
+        elseif a:item['documentation']['kind'] == 'plaintext' " field is MarkupContent
+            let l:completion['info']['value'] .= a:item['documentation']
         endif
     endif
 
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -251,8 +251,8 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
         if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
         elseif type(a:item['documentation']) == type({}) && 
-                    \ a:item['documentation']['kind'] == 'plaintext'
-            " field is MarkupContent
+                    \ has_key(a:item['documentation'], 'value')
+            " field is MarkupContent (string or markdown, we don't care)
             let l:completion['info']['value'] .= a:item['documentation']
         endif
     endif


autoload/lsp/omni.vim
```diff
@@ -248,8 +248,10 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     endif
 
     if has_key(a:item, 'documentation')
-        if type(a:item['documentation']) == type('')
+        if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
+        elseif a:item['documentation']['kind'] == 'plaintext' " field is MarkupContent
```
thomasfaingnaert: Should check for type as well: `type(a:item['documentation']) == type({})`.
Are we going to do something about Markdown as well?

clason: What is the purpose of the check? That it's non-empty? Maybe try to catch that first?

Re Markdown: How would we do that? If it's just formatted text, we can pass that through (in which case we don't need the check), but if it's an embedded image, I don't know how this could be handled in nvim...

thomasfaingnaert: If documentation is neither a string, nor a dict, then the subscript `['kind']` will fail (could be the case if the server returns null for example).

For markdown: I'm just thinking that maybe some servers return some text, but wrap everything in markdown, because in VSCode it won't matter.

clason: True, but the server should check for Markdown capabilities first (texlab does, at least for hover).

But the worst that can happen is a preview window full of an ascii-coded image link ;)

thomasfaingnaert: Ah right, that's another separate capability. I always forget about those :(
Honestly, I'm fine with it either way: whether you decide to check for `plaintext` or not.

An ascii coded image does sound amazing tho :p

clason: Let's be permissive; it's always easier to open (and act on) an issue complaining about weird text in the preview than about something not showing that should.

autoload/lsp/omni.vim
```diff
@@ -248,8 +248,12 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     endif
 
     if has_key(a:item, 'documentation')
-        if type(a:item['documentation']) == type('')
+        if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
+        elseif type(a:item['documentation']) == type({}) && 
+                    \ has_key(a:item['documentation'], 'value')
+            " field is MarkupContent (string or markdown, we don't care)
+            let l:completion['info']['value'] .= a:item['documentation']
```
thomasfaingnaert: Will this not fail because 'info' is a string?

thomasfaingnaert: Ah, I think you want to flip the `value` to the other side?

clason: :man_facepalming:

thomasfaingnaert:
Does this need any additional client capabilities to be advertised?

clason:
No, Markdown support is only tested for `hover` and `signatureHelp`. I guess this means that even without advertising support, clients should expect to be served `MarkupContent`.

thomasfaingnaert:
What about `TextDocumentClientCapabilities.completion.completionItem.documentationFormat`?

clason:
Ah, I missed that. Should we advertise `plaintext` preference and hope the server respects that?

thomasfaingnaert:
Sounds good. Although, technically you could support (a limited subset of) markdown as well.
I suppose quite a few language servers will use markdown?

autoload/lsp.vim
```diff
@@ -386,6 +386,9 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   },
     \   'textDocument': {
     \       'completion': {
+    \           'completionItem': {
+    \              'documentationFormat': 'plaintext'
```
thomasfaingnaert: Think this needs to be array.
```suggestion
    \              'documentationFormat': ['plaintext']
```

thomasfaingnaert:
I just thought of something: if we were to save the 'documentation' field as we get it from the server in the completion item's `info`, we can reuse the logic in `output.vim` for markdown right?

clason:
I don't know of any -- texlab and pyls don't; I'm checking Julia now (that one's always slow to start up).

> I just thought of something: if we were to save the 'documentation' field as we get it from the server in the completion item's info, we can reuse the logic in output.vim for markdown right?

Maybe, but would this do more than stripping the markdown to get the plain text -- which we also get if we ask for plaintext in the first place? (honest question)

thomasfaingnaert:
You would get the syntax highlighting, conceals, etc. from #441 in the documentation popup.

thomasfaingnaert:
For example, for `typescript`, the hover looks like this:
image
I imagine that the documentation for `foo` will look similar, but I haven't tested it yet.

thomasfaingnaert:
It does in VSCode, at least:
image

clason:
Ah, I see, you meant `lsp-hover.vim`. Yes, that would be a good idea. But getting the highlighting would mean additional changes to the output routine, I think. And `pyls` definitely only serves up a string for `documentation`.

thomasfaingnaert:
> Ah, I see, you meant `lsp-hover.vim`. Yes, that would be a good idea. But getting the highlighting would mean additional changes to the output routine, I think.

Would it? For hover, LSP supports MarkedString (special case: normal string), MarkedString[] and MarkupContent.
The documentation of a completion item is string or MarkupContent. So all cases are already implemented.

For hover, we directly pass the response to `...#output()`, so doing this for documentation as well should be fairly straightforward I would think.

clason:
Oh, I think I know what you mean -- instead of just assigning `documentation` to `info`, we'd pass it through the output routine first. Yes, that could work, but a) I can't find the `output()` routine you mentioned and b) I can't test it with the servers I have.

thomasfaingnaert:
Jup, that's what I meant. May not have made that too clear. We can reuse the hover logic and not worry about markdown at all.

> a) I can't find the output() routine you mentioned
Oops, I meant `preview()` in `output.vim` (https://github.com/prabirshrestha/vim-lsp/blob/34397f0b5f72c058c6806cff88f82d055958146c/autoload/lsp/ui/vim/output.vim#L287).
You can see that hover just sends the hover's content directly: https://github.com/prabirshrestha/vim-lsp/blob/34397f0b5f72c058c6806cff88f82d055958146c/autoload/lsp/ui/vim/hover.vim#L38. We can do the same for documentation without too much hassle.

> b) I can't test it with the servers I have.

If you're not opposed to installing node, you can test it with typescript language server:
```
sudo zypper install node npm
sudo npm install -g typescript-language-server
sudo npm install -g typescript
```
(not sure about the package names, I'm on a `deb` distro; it's been some time since I last used SUSE).

autoload/lsp.vim
```diff
@@ -386,6 +386,9 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   },
     \   'textDocument': {
     \       'completion': {
+    \           'completionItem': {
+    \              'documentationFormat': ['plaintext']
```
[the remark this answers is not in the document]
clason: That's the question we're just discussing -- whether we can do something nicer with markdown than stripping it down to plaintext, which is what we're currently asking the server to do.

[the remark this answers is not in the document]
thomasfaingnaert: The answer will most likely be yes.

clason:
@thomasfaingnaert Yes, but currently we're not displaying the documentation at all -- we're relying on `completeopt+=preview` to pick up on the `info` field and display it using the built-in mechanism. Changing that would require more intrusive changes, unless I'm overlooking something obvious?

thomasfaingnaert:
Ah I thought you were going to implement displaying the hover in vim-lsp itself using `...#output#preview(...)`.
Yes, so if we rely on built-in `completeopt+=popup` (does Neovim have that?) we have to use plain-text unfortunately. I don't think there's an easy way to do custom highlights using that built-in functionality?

I think if we want syntax highlight and conceal in documentation, we need to implement the display ourselves and hook into the `CompletionSelectionChanged` autocmd (if there is something like that).

@prabirshrestha Thoughts?

clason:
Which is exactly what #304 is about :) I was not volunteering for implementing that quite yet, sorry if I gave the wrong impression on this PR -- this was just meant as a stopgap to get Documentation for texlab at all.

> Yes, so if we rely on built-in completeopt+=popup (does Neovim have that?) 

No, but https://github.com/ncm2/float-preview.nvim implements something like it (partially in Python, though).

thomasfaingnaert:
#304 is just for lazily fetching the documentation, not necessarily a documentation popup with all bells and whistles :p That will take quite some effort though, I'm fine with only enabling plaintext for now and letting `completeopt` handle it.

thomasfaingnaert:
> If I should still pass through markdown in addition to -- or in preference of? -- plaintext in this PR, I'm fine with making that change. (It's not set in stone, anyway.)

I think I'd rather have plaintext for now, than no-syntax highlighted markdown. I'll let @prabirshrestha decide.

> I guess I got distracted by @prabirshrestha's GIFs showing shiny popups ;)

Who wouldn't :)

thomasfaingnaert:
@clason Does `texlab` use `completionItem/resolve` or is the documentation field populated on `textDocument/completion` already?

thomasfaingnaert:
@clason 
Ah perfect. Should I take a look at `resolve` or is the current behaviour of documentation (with this PR) good enough for now?

clason:
Is `resolve` working now? I thought that was still `todo`. (Even with this (tiny) PR, I don't get citation documentation.)

autoload/lsp.vim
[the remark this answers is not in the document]
thomasfaingnaert: @prabirshrestha For now, we'll probably rely on `completeopt+=popup`, so we have to use plaintext.
I think we probably should leave it as plaintext, and when we call `lsp#ui#vim#output#preview` ourselves, we can add markdown.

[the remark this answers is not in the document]
clason: @thomasfaingnaert you mean `preview`, right? There's no `popup`, either in vim or neovim.

thomasfaingnaert:
It's not working, no. Guess I'll take a look at it then.

clason:
Sounds like a larger task; not too much of a priority.

autoload/lsp.vim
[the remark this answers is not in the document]
thomasfaingnaert: @clason There is actually: https://github.com/vim/vim/blob/840d16fd36bfd1a9fac8200e3dc016b1e3f9c328/runtime/doc/options.txt#L1916-L1920. But `preview` works too. (Though I find it too distracting)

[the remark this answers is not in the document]
thomasfaingnaert: It kind of looks nice if you ignore the colours and weird positioning:
Peek 2019-09-11 22-43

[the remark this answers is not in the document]
clason: Oh, that's _very_ recent! I stand corrected. (It's a standard pattern: A feature request is closed or ignored, and years later silently implemented...)

I consider `preview` a reference implementation to see if it works, but not really useful for actual work. Much too slow, and to quote roxma from ncm2/float-preview.nvim: "it simply pumps text out of my eye spot".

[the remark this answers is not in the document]
thomasfaingnaert: Seems like you can disable the border. Looking a lot better now. Not sure if there's an autocmd that we can hook into for the markdown. I'll have to play around with it a bit.
Peek 2019-09-11 22-50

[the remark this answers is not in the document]
clason: I've opened an issue for neovim: https://github.com/neovim/neovim/issues/10996

clason:
@prabirshrestha Can this be merged? It's a small, self-contained change that strictly adds functionality; all other issues discussed in the comments can (and should) be deferred to separate PRs.

clason:
@prabirshrestha bump?

clason:
@thomasfaingnaert Is there something still missing from this PR?

thomasfaingnaert:
@clason Looks good to me, I'll merge this in.

As an aside, are you using the Julia language server? It seems that one of the dependencies has made a breaking change, resulting in `LanguageServer.jl` not working anymore.

clason:
@thomasfaingnaert Thanks!

> As an aside, are you using the Julia language server? It seems that one of the dependencies has made a breaking change, resulting in LanguageServer.jl not working anymore.

Not really; my understanding is that it isn't quite ready for production (problems keeping up with Julia development -- it only got full 1.0 support a month ago). It also takes too long to spin up, especially after updates. 

Are you using https://github.com/machakann/vim-lsp-julia?

thomasfaingnaert:
> Are you using https://github.com/machakann/vim-lsp-julia?

No, I registered it myself using
```viml
            \ 'name': 'julia',
            \ 'cmd': {server_info->['julia', '--startup-file=no', '--history-file=no', '-e', '
            \   using LanguageServer;
            \   using Pkg;
            \   import StaticLint;
            \   import SymbolServer;
            \   env_path = dirname(Pkg.Types.Context().env.project_file);
            \   debug = false;
            \
            \   server = LanguageServer.LanguageServerInstance(stdin, stdout, debug, env_path, "", Dict());
            \   server.runlinter = true;
            \   run(server);
            \   ']},
            \ 'whitelist': ['julia'],
```
but it seems that's pretty much the same as what https://github.com/machakann/vim-lsp-julia does.

clason:
I think the recommendation is still to `add LanguageServer#master` (and similarly for the dependencies), even if it's no longer the official installation instruction.

thomasfaingnaert:
Doesn't seem to work that well either; I guess I'll wait until it becomes a bit more stable.

clason:
Yeah, there are several issues on "LanguageServer crashing" that aren't resolved yet...

PR #502, opened 2019-09-10
Avoid wrong buffer open

hrsh7th:
Avoiding wrong buffer sending to server.

I met problems when open and change `typescript` buffer and swich to other `not typescript` buffer quickly. (`not typescript` buffer will be red all lines)

The problem's reason is sending "not typescript" buffer as `typescript` source.

I think it occurs in following flow.

1. Open typescript buffer and changes.
2. Run `s:on_text_document_did_change()` by autocmd.
    - `bufnr('%')` is save to variable... `A`
3. After timer tick, call `s:ensure_flush`.
4. Call `lsp#ui#vim#folding#send_request` in `s:ensure_flush`.
    - Somtimes `bufnr('%')` are not same for `A`.

The solution of this PR is adding `bufnr` argument for `lsp#send_request`.

I worry to change `lsp#send_request` it is very common api...

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -782,11 +782,12 @@ function! s:get_versioned_text_document_identifier(buf, buffer_info) abort
 endfunction
 
 function! lsp#send_request(server_name, request) abort
+    let l:bufnr = get(a:request, 'bufnr', bufnr('%'))
     let l:Cb = has_key(a:request, 'on_notification') ? a:request['on_notification'] : function('s:Noop')
     let l:request = copy(a:request)
     let l:request['on_notification'] = {id, data, event->l:Cb(data)}
     call lsp#utils#step#start([
-        \ {s->s:ensure_flush(bufnr('%'), a:server_name, s.callback)},
+        \ {s->s:ensure_flush(l:bufnr, a:server_name, s.callback)},
         \ {s->s:is_step_error(s) ? l:Cb(s.result[0]) : s:send_request(a:server_name, l:request) },
         \ ])
 endfunction
--- a/autoload/lsp/ui/vim/folding.vim
+++ b/autoload/lsp/ui/vim/folding.vim
@@ -61,7 +61,8 @@ function! lsp#ui#vim#folding#send_request(server_name, buf, sync) abort
                 \   'textDocument': lsp#get_text_document_identifier(a:buf)
                 \ },
                 \ 'on_notification': function('s:handle_fold_request', [a:server_name]),
-                \ 'sync': a:sync
+                \ 'sync': a:sync,
+                \ 'bufnr': a:buf
                 \ })
 endfunction
 


mattn:
You should add the last argument for current buffer for all part that use `sync': a:sync`.

hrsh7th:
My understanding is that `a:sync` is unrelated this problem.

Following code has similar problem.

```VimL
let l:bufnr = bufnr('%')
let l:has_suitable_server = lsp#get_whitelisted_servers(l:bufnr)
if l:has_suitable_server
  call timer_start(1000, { _ -> lsp#send_request(...) })
endif
```

It checks `lsp#get_whitelisted_servers` before call `lsp#send_request`.

But call `lsp#send_request` is after timer tick and `lsp#send_request` uses `bufnr('%')` (It's wrong, Because sometimes it `bufnr('%')` is changed by user).

So currently `lsp#send_request` must call in synchronous context after capability checks, I think.

This PR aims to able to specifying `bufnr` explicitly for `lsp#send_request`.

Change the code to following after this PR and It's OK.

```VimL
let l:bufnr = bufnr('%')
let l:has_suitable_server = lsp#get_whitelisted_servers(l:bufnr)
if l:has_suitable_server
  call timer_start(1000, { _ -> lsp#send_request(..., l:bufnr) })
endif
```

I checked `vim-lsp`'s codes thats are mostly OK.

https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/ui/vim.vim#L541 has only possibility wrong.
I will be support this point if this solution are acceptable.

prabirshrestha:
@mattn @hrsh7th Any update on this? I'm seeing "cannot add text property to unloaded buffer" occasionally.

hrsh7th:
@prabirshrestha
Occured the error in this PR?

Currently textprop is used in diagnostics and folding only.

I guess 'cannot add text property' was caused from diagnostics side.

Probably this PR solves the textprop error also.

If `no-typescript` buffer sent to the server, the client will receive diagnostics(for all lines).

vim-lsp will set the diagnostic to textprop if enabled.

If `no-typescript` buffer is `plugin's buffer` like `denite` `quickpick`, the buffer often unloaded (And This is often done quickly).

So this PR solves the problem maybe,because this PR avoid sending wrong buffer to the server.

prabirshrestha:
Error is from master without this PR.

hrsh7th:
Thanks for information.

Very sorry for my poor english.
I can't expaining this problem in english.

I try to expain the problem.

```VimL
s:on_text_document_did_change " capture bufnr('%') to `l:buf`.

... async ...

s:ensure_flush " call with the `l:buf`.
  s:ensure_start
  s:ensure_init
  s:ensure_conf
  s:ensure_open
    lsp#ui#vim#folding#send_request(a:server_name, a:buf, 0)
      lsp#send_request " NG! used not `l:buf` but `bufnr('%').
  s:ensure_changed
    lsp#ui#vim#folding#send_request(a:server_name, a:buf, 0)
      lsp#send_request " NG! used not `l:buf` but `bufnr('%').
```

hrsh7th:
I will try to create repro steps.
I think this problem is high priority.

and I will change the signature to include new 'bufnr' argument to the 'send_request' params.

habamax:
I have tried it and it feels strange.

1. None of my errors were highlighted (not in ruby, not in vim script) -- So not sure if it is fixed or not.
2. :LSPHover opens documentation in a split, not in popup as master does
3. :LSPDefinition doesn't work (No definition found), with master it does.

hrsh7th:
Oh... thanks. I check more deeply.

habamax:
@hrsh7th sorry! I was checking your master not the branch!

It works ok, as no other buffer is checked!

image

habamax:
The code change looks good, @prabirshrestha what should be done to accept this PR?

prabirshrestha:
Merged. Thanks.

PR #504, opened 2019-09-11
Fix diagnostic highlighting on incorrect texts (#503)

### LINKED ISSUE #503: highlight on incorrect texts
This issue hasn't happened before, only with recent commits, I'm not sure which one causes this.

In my case, I set:

```
hi link LspErrorHighlight Underlined
hi link LspWarningHighlight Underlined
hi link LspInformationHighlight Underlined
hi link LspHintHighlight Underlined
```

Notice these underline highlights are on incorrect texts, which that should be on the ones above.

Screen Shot 2019-09-11 at 14 57 58

thenewvu:
Fixes #503

### DIFF
--- a/autoload/lsp/ui/vim/highlights.vim
+++ b/autoload/lsp/ui/vim/highlights.vim
@@ -103,7 +103,7 @@ function! s:place_highlights(server_name, path, diagnostics) abort
 
             let l:name = get(s:severity_sign_names_mapping, l:item['severity'], 'LspError')
             let l:hl_name = l:name . 'Highlight'
-            call nvim_buf_add_highlight(l:bufnr, l:ns, l:hl_name, l:line, l:start_col, l:end_col)
+            call nvim_buf_add_highlight(l:bufnr, l:ns, l:hl_name, l:line - 1, l:start_col - 1, l:end_col - 1)
         endfor
     endif
 endfunction


clason:
By the way: If you put `Fixes #503` in the description, the PR is linked to the issue, and the issue will automatically be closed when the PR is merged.

autoload/lsp/ui/vim/highlights.vim
```diff
@@ -95,11 +95,11 @@ function! s:place_highlights(server_name, path, diagnostics) abort
 
     if !empty(a:diagnostics) && l:bufnr >= 0
         for l:item in a:diagnostics
-            let l:line = l:item['range']['start']['line'] + 1
+            let l:line = l:item['range']['start']['line']
```
clason:
```suggestion
            let l:line = l:item['range']['start']['line'] + 1
```

clason: I can't suggest the change from here, but instead you need to pass `l:line-1` to `nvim_buf_add_highlight` in line 106.

clason:
The issue is that the line number is used by `to_col` to compute the correct offset in bytes, and it expects the lines to be 1-indexed (that's why @machakann changed that in their PR). In contrast, `nvim_buf_add_highlight` expects 0-indexed lines.

thenewvu:
@clason applied your suggestion.

clason:
Nitpick: Maybe it's cleaner to put all the `-1` in the call to `nvim_buff_add_highlight` so that it's more obvious that we're correcting for 0-indexing here?

thenewvu:
> Nitpick: Maybe it's cleaner to put all the `-1` in the call to `nvim_buff_add_highlight` so that it's more obvious that we're correcting for 0-indexing here?

Done, I don't know what I'm doing anyway ;)

clason:
You seem to be doing it quite successfully, though :) Thanks for catching that! I missed it when testing the PR with this commit.

machakann:
It's my bad. Thank you, I think this patch is correct.

thomasfaingnaert:
Merged, thanks to @thenewvu for fix and @clason for reviewing.

PR #507, opened 2019-09-12
[WIP] Popup documentation

thomasfaingnaert:
Uses the hover logic to display documentation in the popup/floating window (like `completeopt+=popup`, but with Markdown capabilities).

Vim:
vim

Neovim:
nvim

TODO:
- [x] Cleanup Neovim window
- [x] Neovim positioning
- [ ] Use `completeopt+=popup` positioning logic for Vim (?)
- [ ] Allow disabling

PR #509, opened 2019-09-13
[WIP] Refactor neovim floats

clason:
The purpose of this PR is to avoid entering the buffer opened with `nvim_open_win`, which triggers -- possibly very expensive -- `CursorMoved` autocommands on the original buffer, see #507.

This PR also contains commits from #500 and #507 in order to easily test the `Documentation` behavior; it should be removed before merging.

This is not quite working yet:
1. Documentation preview contents are right-aligned, so that the beginning rather than the end of long lines are truncated. (That's an issue with `documentation.vim` trying to put the float on the left side of the pum but failing for some reason.)
2. Hover doesn't show if triggered on the first line of the buffer.
3. PeekDefinition doesn't show for if the target range is too large, and if the buffer contains unsaved changes, these are not picked up.

@jerdna-regeiz @thomasfaingnaert I would appreciate feedback, since the interaction of the functions in `output.vim` isn't quite clear to me.

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -288,16 +288,16 @@ function! s:align_preview(options) abort
     endif
 endfunction
 
-function! lsp#ui#vim#output#get_size_info() abort
+function! lsp#ui#vim#output#get_size_info(lines) abort
     " Get size information while still having the buffer active
-    let l:maxwidth = max(map(getline(1, '$'), 'strdisplaywidth(v:val)'))
+    let l:maxwidth = max(map(a:lines, 'strdisplaywidth(v:val)'))
     if g:lsp_preview_max_width > 0
       let l:bufferlines = 0
       let l:maxwidth = min([g:lsp_preview_max_width, l:maxwidth])
 
       " Determine, for each line, how many "virtual" lines it spans, and add
       " these together for all lines in the buffer
-      for l:line in getline(1, '$')
+      for l:line in a:lines
         let l:num_lines = str2nr(string(ceil(strdisplaywidth(l:line) * 1.0 / g:lsp_preview_max_width)))
         let l:bufferlines += max([l:num_lines, 1])
       endfor
@@ -344,7 +344,7 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
     call setbufvar(winbufnr(s:winid), 'lsp_do_conceal', l:do_conceal)
     call lsp#ui#vim#output#setcontent(s:winid, l:lines, l:ft)
 
-    let [l:bufferlines, l:maxwidth] = lsp#ui#vim#output#get_size_info()
+    let [l:bufferlines, l:maxwidth] = lsp#ui#vim#output#get_size_info(getlines(1, '$'))
 
     if s:use_preview
         " Set statusline
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -143,10 +143,10 @@ function! lsp#ui#vim#output#floatingpreview(data) abort
   return s:winid
 endfunction
 
-function! s:setcontent(lines, ft) abort
+function! lsp#ui#vim#output#setcontent(winid, lines, ft) abort
   if s:use_vim_popup
     " vim popup
-    call setbufline(winbufnr(s:winid), 1, a:lines)
+    call setbufline(winbufnr(a:winid), 1, a:lines)
     let l:lightline_toggle = v:false
     if exists('#lightline') && !has('nvim')
       " Lightline does not work in popups but does not recognize it yet.
@@ -154,7 +154,7 @@ function! s:setcontent(lines, ft) abort
       let l:lightline_toggle = v:true
       call lightline#disable()
     endif
-    call win_execute(s:winid, 'setlocal filetype=' . a:ft . '.lsp-hover')
+    call win_execute(a:winid, 'setlocal filetype=' . a:ft . '.lsp-hover')
     if l:lightline_toggle
       call lightline#enable()
     endif
@@ -167,7 +167,7 @@ function! s:setcontent(lines, ft) abort
   endif
 endfunction
 
-function! s:adjust_float_placement(bufferlines, maxwidth) abort
+function! lsp#ui#vim#output#adjust_float_placement(bufferlines, maxwidth) abort
     if s:use_nvim_float
       let l:win_config = {}
       let l:height = min([winheight(s:winid), a:bufferlines])
@@ -284,6 +284,26 @@ function! s:align_preview(options) abort
     endif
 endfunction
 
+function! lsp#ui#vim#output#get_size_info() abort
+    " Get size information while still having the buffer active
+    let l:maxwidth = max(map(getline(1, '$'), 'strdisplaywidth(v:val)'))
+    if g:lsp_preview_max_width > 0
+      let l:bufferlines = 0
+      let l:maxwidth = min([g:lsp_preview_max_width, l:maxwidth])
+
+      " Determine, for each line, how many "virtual" lines it spans, and add
+      " these together for all lines in the buffer
+      for l:line in getline(1, '$')
+        let l:num_lines = str2nr(string(ceil(strdisplaywidth(l:line) * 1.0 / g:lsp_preview_max_width)))
+        let l:bufferlines += max([l:num_lines, 1])
+      endfor
+    else
+      let l:bufferlines = line('$')
+    endif
+
+    return [l:bufferlines, l:maxwidth]
+endfunction
+
 function! lsp#ui#vim#output#preview(server, data, options) abort
     if s:winid && type(s:preview_data) == type(a:data)
        \ && s:preview_data == a:data
@@ -303,7 +323,7 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
     let s:preview_data = a:data
     let l:lines = []
     let l:syntax_lines = []
-    let l:ft = s:append(a:data, l:lines, l:syntax_lines)
+    let l:ft = lsp#ui#vim#output#append(a:data, l:lines, l:syntax_lines)
 
     if has_key(a:options, 'filetype')
         let l:ft = a:options['filetype']
@@ -318,23 +338,9 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
 
     call setbufvar(winbufnr(s:winid), 'lsp_syntax_highlights', l:syntax_lines)
     call setbufvar(winbufnr(s:winid), 'lsp_do_conceal', l:do_conceal)
-    call s:setcontent(l:lines, l:ft)
+    call lsp#ui#vim#output#setcontent(s:winid, l:lines, l:ft)
 
-    " Get size information while still having the buffer active
-    let l:maxwidth = max(map(getline(1, '$'), 'strdisplaywidth(v:val)'))
-    if g:lsp_preview_max_width > 0
-      let l:bufferlines = 0
-      let l:maxwidth = min([g:lsp_preview_max_width, l:maxwidth])
-
-      " Determine, for each line, how many "virtual" lines it spans, and add
-      " these together for all lines in the buffer
-      for l:line in getline(1, '$')
-        let l:num_lines = str2nr(string(ceil(strdisplaywidth(l:line) * 1.0 / g:lsp_preview_max_width)))
-        let l:bufferlines += max([l:num_lines, 1])
-      endfor
-    else
-      let l:bufferlines = line('$')
-    endif
+    let [l:bufferlines, l:maxwidth] = lsp#ui#vim#output#get_size_info()
 
     if s:use_preview
         " Set statusline
@@ -353,7 +359,7 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
     if s:winid && (s:use_vim_popup || s:use_nvim_float)
       if s:use_nvim_float
         " Neovim floats
-        call s:adjust_float_placement(l:bufferlines, l:maxwidth)
+        call lsp#ui#vim#output#adjust_float_placement(l:bufferlines, l:maxwidth)
         call s:set_cursor(l:current_window_id, a:options)
         call s:add_float_closing_hooks()
       elseif s:use_vim_popup
@@ -371,10 +377,10 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
     return ''
 endfunction
 
-function! s:append(data, lines, syntax_lines) abort
+function! lsp#ui#vim#output#append(data, lines, syntax_lines) abort
     if type(a:data) == type([])
         for l:entry in a:data
-            call s:append(entry, a:lines, a:syntax_lines)
+            call lsp#ui#vim#output#append(entry, a:lines, a:syntax_lines)
         endfor
 
         return 'markdown'
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -344,7 +344,7 @@ function! lsp#ui#vim#output#preview(server, data, options) abort
     call setbufvar(winbufnr(s:winid), 'lsp_do_conceal', l:do_conceal)
     call lsp#ui#vim#output#setcontent(s:winid, l:lines, l:ft)
 
-    let [l:bufferlines, l:maxwidth] = lsp#ui#vim#output#get_size_info(getlines(1, '$'))
+    let [l:bufferlines, l:maxwidth] = lsp#ui#vim#output#get_size_info(getline(1, '$'))
 
     if s:use_preview
         " Set statusline
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -112,15 +112,8 @@ function! lsp#ui#vim#output#floatingpreview(data) abort
     endif
 
     let l:opts = s:get_float_positioning(l:height, l:width)
-
-    let s:winid = nvim_open_win(buf, v:true, l:opts)
-    call nvim_win_set_option(s:winid, 'winhl', 'Normal:Pmenu,NormalNC:Pmenu')
-    call nvim_win_set_option(s:winid, 'foldenable', v:false)
-    call nvim_win_set_option(s:winid, 'wrap', v:true)
-    call nvim_win_set_option(s:winid, 'statusline', '')
-    call nvim_win_set_option(s:winid, 'number', v:false)
-    call nvim_win_set_option(s:winid, 'relativenumber', v:false)
-    call nvim_win_set_option(s:winid, 'cursorline', v:false)
+    let l:opts['style'] = 'minimal'
+    let s:winid = nvim_open_win(buf, v:false, l:opts)
     " Enable closing the preview with esc, but map only in the scratch buffer
     nmap <buffer><silent> <esc> :pclose<cr>
   elseif s:use_vim_popup
@@ -144,9 +137,9 @@ function! lsp#ui#vim#output#floatingpreview(data) abort
 endfunction
 
 function! lsp#ui#vim#output#setcontent(winid, lines, ft) abort
-  if s:use_vim_popup
-    " vim popup
-    call setbufline(winbufnr(a:winid), 1, a:lines)
+  if s:use_vim_popup || s:use_nvim_float
+    let l:buf = winbufnr(a:winid)
+    call setbufline(l:buf, 1, a:lines)
     let l:lightline_toggle = v:false
     if exists('#lightline') && !has('nvim')
       " Lightline does not work in popups but does not recognize it yet.
@@ -154,12 +147,23 @@ function! lsp#ui#vim#output#setcontent(winid, lines, ft) abort
       let l:lightline_toggle = v:true
       call lightline#disable()
     endif
-    call win_execute(a:winid, 'setlocal filetype=' . a:ft . '.lsp-hover')
+    
+    let l:filetype = a:ft . '.lsp-hover'
+    if s:use_vim_popup
+      call win_execute(a:winid, 'setlocal filetype=' . l:filetype)
+    elseif s:use_nvim_float
+      let l:height = len(a:lines)
+      let l:width = max(map(a:lines, 'strdisplaywidth(v:val)'))
+      let l:opts = s:get_float_positioning(l:height, l:width)
+      call nvim_win_set_config(a:winid, l:opts)
+      call nvim_buf_set_option(l:buf, 'filetype', l:filetype)
+    endif
+
     if l:lightline_toggle
       call lightline#enable()
     endif
   else
-    " nvim floating or preview
+    " preview
     call setline(1, a:lines)
 
     setlocal readonly nomodifiable
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -386,6 +386,9 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \   },
     \   'textDocument': {
     \       'completion': {
+    \           'completionItem': {
+    \              'documentationFormat': ['plaintext']
+    \           },
     \           'completionItemKind': {
     \              'valueSet': lsp#omni#get_completion_item_kinds()
     \           }
@@ -695,6 +698,8 @@ function! s:handle_initialize(server_name, data) abort
     if g:lsp_signature_help_enabled
         call lsp#ui#vim#signature_help#setup()
     endif
+
+    call lsp#ui#vim#documentation#setup()
 
     doautocmd User lsp_server_init
 endfunction
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -248,8 +248,12 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     endif
 
     if has_key(a:item, 'documentation')
-        if type(a:item['documentation']) == type('')
+        if type(a:item['documentation']) == type('') " field is string
             let l:completion['info'] .= a:item['documentation']
+        elseif type(a:item['documentation']) == type({}) && 
+                    \ has_key(a:item['documentation'], 'value')
+            " field is MarkupContent (hopefully 'plaintext')
+            let l:completion['info'] .= a:item['documentation']['value']
         endif
     endif
 
--- a/autoload/lsp/ui/vim/documentation.vim
+++ b/autoload/lsp/ui/vim/documentation.vim
@@ -1,0 +1,75 @@
+let s:use_vim_popup = has('patch-8.1.1517') && !has('nvim')
+let s:use_nvim_float = exists('*nvim_open_win') && has('nvim')
+
+let s:last_popup_id = -1
+
+function! s:complete_done() abort
+    " Use a timer to avoid textlock (see :h textlock).
+    let l:event = deepcopy(v:event)
+    call timer_start(0, {-> s:show_documentation(l:event)})
+endfunction
+
+function! s:show_documentation(event) abort
+    call s:close_popup()
+
+    if !has_key(a:event['completed_item'], 'info') || empty(a:event['completed_item']['info'])
+        return
+    endif
+
+    let l:right = wincol() < winwidth(0) / 2
+
+    " TODO: Neovim
+    if l:right
+        let l:line = a:event['row'] + 1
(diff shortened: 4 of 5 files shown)


autoload/lsp/ui/vim/output.vim
```diff
@@ -112,15 +112,8 @@ function! lsp#ui#vim#output#floatingpreview(data) abort
...
-    call nvim_win_set_option(s:winid, 'foldenable', v:false)
-    call nvim_win_set_option(s:winid, 'wrap', v:true)
-    call nvim_win_set_option(s:winid, 'statusline', '')
-    call nvim_win_set_option(s:winid, 'number', v:false)
-    call nvim_win_set_option(s:winid, 'relativenumber', v:false)
-    call nvim_win_set_option(s:winid, 'cursorline', v:false)
+    let l:opts['style'] = 'minimal'
+    let s:winid = nvim_open_win(buf, v:false, l:opts)
     " Enable closing the preview with esc, but map only in the scratch buffer
     nmap <buffer><silent> <esc> :pclose<cr>
```
thomasfaingnaert: Now that we're not focusing the float, this would apply to the main buffer, I believe.

clason: Ah, you're right. Didn't notice that. How weird...

thomasfaingnaert:
> Hover doesn't show if triggered on the first line of the buffer.

What do you mean by this? It seems to work fine on my end:
image

thomasfaingnaert:
Maybe we should also consider getting rid of the "adjust size" functions, and only have a function "set content" that sets the content, and sets the size automatically (just like Vim floats behave).

clason:
Weird, doesn't work for me (with texlab).

clason:
Yeah, there's too many intersecting code paths currently depending on the LSP function. I don't know the difference between half of these functions... Might be good to modularize it.

thomasfaingnaert:
Yeah, might be a good idea to cleanup `output.vim` a bit before we rely on it for documentation and other features.

thomasfaingnaert:
Would you be OK with turning this PR is a general "refactor output.vim" PR?
We can have a discussion on how to API should look like here.

clason:
I would be OK with that; unless there is someone better suited for actually implementing it? (Is there a way to allow others to push to this branch?)

thomasfaingnaert:
You can also add them as collaborators to your fork, but then they'll be able to push to all branches, which may or may not be what you want.

clason:
Sounds good. This PR probably ends up being junked anyway and replaced with a clean proposal for merging.

thomasfaingnaert:
Things would be a lot simpler if nvim floats mirrored Vim's API somewhat (or vice versa) ;)

clason:
I trust you ;) Maybe that's more convenient.

thomasfaingnaert:
Probably good though to precisely document which functions will be available before starting, and what their parameters should be.

For the creation of the buffer, does it make sense to have that in the common API, because all features will have a different way of doing that: hover/peek def will want to create one located at the cursor, documentation on Vim will want to use the Vim float (and not open one itself), documentation on Neovim will create one manually placed next to the completion menu. (At least until `completeopt+=popup` is merged).

clason:
One possibility for the API would be to first define a lower-level API that mirrors what you would like to have, and then for vim just pass it through while for neovim we'd try to implement the same behavior with their (even lower-level) API -- if they end up including higher-level APIs, too, we then just replace them.

thomasfaingnaert:
As a start (basically what we need from Vim's API):

- `popup_create({what}, {options})`: Open a popup showing {what} (buffer number, string, string[])
  `options` is a dict, can cointain options from `:h popup_create-arguments` (though only the ones we need)
  Returns the window ID of the float/popup.

For now, I think we need `line`, `col`, `pos`, `maxwidth`, `firstline`, `moved`.

- `popup_close({winid})`: Closes the popup with the given ID. (Do we actually need this? Popups should close when the cursor moves, though I guess we need it for Neovim floats and pressing ESC inside them).

- `poup_set_content({widid}, {data})`: Set the content of the popup with given ID to `{data}`. `{data}` can be anything returned from the language server: String, String[], MarkedString, MarkedString[], MarkupContent


Something like this?

clason:
Probably also `style` (plaintext vs. markdown)?

What about wrapping?

(Neovim plans to add higher-level APIs in the 0.5.0 cycle. One plan is to add functions for "ballooneval-as-floats", which seems to be half of what we need here, the other half being `completeopt`.)

thomasfaingnaert:
Well, for Vim `wrap` is true by default, so I haven't included it. (The idea is that for Neovim, we would implement wrap ourselves if needed)

thomasfaingnaert:
For the `style`: this would be obtained from the `{data}` to `popup_set_content`.

thomasfaingnaert:
Is there an approximate release date for `0.5.0` actually?

clason:
So `wrap` would be default -- is there a situation for which we don't want this?

(No release date for `0.5.0`, they're still working on `0.4.0`)

thomasfaingnaert:
I can't really think of a situation where you _wouldn't_ want to wrap.

thomasfaingnaert:
Let's maybe also discuss the opening and setting content: would it make sense to have the `open` function not take a `{what}` argument, but just do the creation of a buffer, floating window, and positioning (but not sizing) based on its `{options}`?
Then, the sizing would happen in `set_content`.

clason:
I think that's related to the question of whether we want to have a higher level API.

That one usually just says: put this text here, and maybe note that there might come more.

It would be good to separate the implementation as much as possible from the LSP functionality. (Possibly even to the extent of making this into an external library?)

On 13 Sep 2019 20:25, Thomas Faingnaert <notifications@github.com> wrote:

Let's maybe also discuss the opening and setting content: would it make sense to have the open function not take a {what} argument, but just do the creation of a buffer, floating window, and positioning (but not sizing) based on its {options}?
Then, the sizing would happen in set_content.

clason:
Maybe allow for empty `what` if you want to fill (repeatedly) later? And then just bypass all the rest?

Would be indistinguishable from having separate open + fill functions with a thin wrapper on top, though.

On 13 Sep 2019 20:25, Thomas Faingnaert <notifications@github.com> wrote:

Let's maybe also discuss the opening and setting content: would it make sense to have the open function not take a {what} argument, but just do the creation of a buffer, floating window, and positioning (but not sizing) based on its {options}?
Then, the sizing would happen in set_content.

thomasfaingnaert:
> It would be good to separate the implementation as much as possible from the LSP functionality. (Possibly even to the extent of making this into an external library?)

The only LSP specific thing is the `set_content`. I suppose we can limit `set_content` to a `String[]`, then. Though, at that point, it might just make sense to have no `set_content` function, just an `open` function that takes a `String[]`.
The advantage there is that it would likely avoid a lot of sizing problems when updating the buffer contents after it has been displayed?

The "low level" API would then just be "open a float/popup/preview with these lines (String[]), maxwidth, minwidth, positioning, close properties; returning the window ID"; and "close the popup/... with this window ID". (This is the non-LSP specific external library, if you will).

For LSP, we can essentially convert the `MarkedString[]`, `MarkupContent`/... into `String[]` (we already do this anyway), pass this to `low_level_api#create()`, and store the window ID.
To get fancy highlighting, we just do a `win_execute(returned_win_id, 'setlocal filetype=lsp-hover')`, `setbufvar(winbufnr(...), 'lsp_syntax_lines', ...)`.
Also makes sense to move the highlighting logic to the "low level API": `apply_syntax_highlights(winid, 'markdown', [{'line': 1, 'language': 'typescript'}, ...])`.

thomasfaingnaert:
Actually, forgot we still need `set_content` for the documentation, because we don't create the float ourselves in that case. The argument of limiting the content to `String[]` still applies though.

clason:
The idea was to abstract as much as possible from the divergent popup/float/preview implementations; I thought keeping the top level as a thin wrapper dispatching to the respective functions (or two) and the low-level as basically just calling external APIs would help -- but then we'd need the intermediate level to make up for the different ways that the structures are interacted with. (I'm thinking of the very different needs of `Hover` and `Documentation`.) But I might be overthinking this; let's keep it at two levels. (Actually, modest proposal: can we drop preview, at least for now?)

> The "low level" API would then just be "open a float/popup/preview with these lines (String[]), maxwidth, minwidth, positioning, close properties; returning the window ID"; and "close the popup/... with this window ID". (This is the non-LSP specific external library, if you will).

When these functions exist, otherwise we'd have to stitch them together using lower-level APIs ourselves.

> Also makes sense to move the highlighting logic to the "low level API":

Yes.

thomasfaingnaert:
I'm all for dropping preview for a moment. I think the 2 most difficult things in the design will be:
1. the difference between preview on the one hand, and popup/float on the other; and
2. the differentce between hover and peek on the one hand (they're essentially the same), and documentation on the other (there's a even a big difference between Vim and Neovim)

Could you also make a quick overview of how you think the API should look like, might be useful to compare.

clason:
> I'm all for dropping preview for a moment. I think the 2 most difficult things in the design will be:
...

I agree. What I had in mind was (using `tooltip` as generic term for want of a better one):

1. Top-level API (`tooltip.vim`):
a) `show_cursor_tooltip(lines, options)`
called by hover, peek, signature, with options controlling whether it's plaintext, markdown, or syntax highlighted (peek)); positioning is based on `.`, closing is done by autocommands
b) `winid = show_pum_tooltip(winid,lines, position,options)`
called by documentation
c) `update_pum_tooltip(winid,lines,position,options)`

2. Intermediate-level would be backend specific, but with a common signature (and possibly very thin): `create_tooltip`, `set_contents`, `set_position`, `close_tooltip`.

3. Low-level would map these to the backend APIs, introducing wrappers as necessary (that can be removed as the backend API becomes more higher level. (Here I have no idea yet; would depend on what's needed or seems useful when writing 2.)

(2.-3. could be an external plugin, if that seemed useful.)

thomasfaingnaert:
>     1. Top-level API (`tooltip.vim`):
>        a) `show_cursor_tooltip(lines, options)`
>        called by hover, peek, signature, with options controlling whether it's plaintext, markdown, or syntax highlighted (peek)); positioning is based on `.`, closing is done by autocommands

I assume with `.` you mean current cursor position? For the closing: would this be the responsibility of `show_cursor_tooltip` (i.e. close on cursor move), or would the caller explicitly need to register autocmds? (For Vim popups, you can specify if you want to close it on cursor move or not; might be worth making this an option as well. I assume you want hover and peek to close on cursor move, but not signature).

>        b) `winid = show_pum_tooltip(winid,lines, position,options)`
>        called by documentation

If this opens the pum tooltip, you don't need the first argument :)

>        c) `update_pum_tooltip(winid,lines,position,options)`
>

>     2. Intermediate-level would be backend specific, but with a common signature (and possibly very thin): `create_tooltip`, `set_tooltip_contents`, `set_tooltip_position`, `close_tooltip`. Would also include "utility functions" for syntax highlighting, markdown stripping etc.

Am I correct in assuming their responsibilities are:
- `create_tooltip`: creates a hidden floating window, and an empty buffer associated with it, with undefined position
- `set_tooltip_contents`: sets the contents of a buffer (String[] only?), and also unhides it (just to avoid flicker); retriggers size calculation
- `set_tooltip_position`: sets the position of the popup correctly (can be cursor-relative, pum-relative, etc.)

>     3. Low-level would map these to the backend APIs, introducing wrappers as necessary (that can be removed as the backend API becomes more higher level. (Here I have no idea yet; would depend on what's needed or seems useful when writing 2.)

I think we'll have enough with these two levels, but yeah, if the need arises, we can abstract further down.

> (2.-3. could be an external plugin, if that seemed useful.)

Looks good, probably need to think about precisely which arguments each layer accepts I guess?

thomasfaingnaert:
Also; in the top level, to add highlighting to specific lines, we will need an `syntaxlines` argument.
`[{'line':1,'language':'tex'},...]`

thomasfaingnaert:
For the `show_pum_tooltip`, what's the purpose of the `position`?

thomasfaingnaert:
I also think `update_pum_tooltip` can be a generic `update_tooltip` probably (we don't need it right now, but it may be handy I guess).

clason:
> I assume with `.` you mean current cursor position? 

Yes; the specifics depend on the backend. 

By the way, neovim (recently) allows
```viml
call nvim_open_win(l:buf,v:false,{'relative':'cursor','row':0,'col':0,...})
```
to open a float at the cursor. This means you only need to calculate where you want the anchor to be :)

> Also; in the top level, to add highlighting to specific lines, we will need an syntaxlines argument.
[{'line':1,'language':'tex'},...]

Could be combined if we allow `{'line':*, 'language':'tex'}` to specify that the whole buffer is `ft=tex`. Would have to think how to best abstract it away from the top-level.

> For the show_pum_tooltip, what's the purpose of the position?

`left`/`right`, `above`/`below`

> I also think update_pum_tooltip can be a generic update_tooltip probably (we don't need it right now, but it may be handy I guess).

Probably, unless `cursor_tooltips` and `pum_tooltips` have too much differing internal details.

> For the closing: would this be the responsibility of `show_cursor_tooltip` (i.e. close on cursor move), or would the caller explicitly need to register autocmds? (For Vim popups, you can specify if you want to close it on cursor move or not; might be worth making this an option as well. I assume you want hover and peek to close on cursor move, but not signature).

Good question, especially in combination with the "focusing" via double-tapping. I don't know.

Maybe best to make that an option, too?

> If this opens the pum tooltip, you don't need the first argument :)

True :) I changed my mind while typing...

> Am I correct in assuming their responsibilities are:

> * `create_tooltip`: creates a hidden floating window, and an empty buffer associated with it, with undefined position

Yes; if you put it like that, there's not much to do, though...

> * `set_tooltip_contents`: sets the contents of a buffer (String[] only?), and also unhides it (just to avoid flicker); retriggers size calculation

Yes. (What would be the alternative? If Markup: yes, parsing would be higher-level -- unless you think it might make sense to push this down, in which case the function needs more options.)

> * `set_tooltip_position`: sets the position of the popup correctly (can be cursor-relative, pum-relative, etc.)

Yes; possibly including resizing here? (If you move the tooltip to the border, it might have to shrink. Then you'd want to not do this in `set_tooltip_contents`. (All this would be hidden by the top-level API, anyway.)

> I think we'll have enough with these two levels, but yeah, if the need arises, we can abstract further down.

Agreed.

> Looks good, probably need to think about precisely which arguments each layer accepts I guess?

Yes. I would suggest starting top down: Going through the LSP functions and writing the API calls we _want_ to have, and then implementing them using made-up lower-level calls, and then writing them separately and independently for vim and neovim (probably iterating whenever we hit a snag).

thomasfaingnaert:
> Could be combined if we allow {'line':*, 'language':'tex'} to specify that the whole buffer is ft=tex. Would have to think how to best abstract it away from the top-level.

I would split up the filetype of the buffer (which determines global syntax highlighting, ftplugin logic etc) from the syntax highlight of each line.
The way I have implemented it right now, each line is highlighted separately, so you can't continue statements across lines. Probably better to have `{'line_start': 1, 'line_end': 3, 'language': 'tex'}` instead, or would that be too complicated?
(In any case, filetype should be a separate setting IMO, because it also includes ftplugin, not only syntax.)

clason:
Yes, that makes sense, although I'd probably mirror the LSP structure with `{'range':{'start':...,'end':...}`.

thomasfaingnaert:
> By the way, neovim (recently) allows
> 
> ```viml
> call nvim_open_win(l:buf,v:false,{'relative':'cursor','row':0,'col':0,...})
> ```
> 
> to open a float at the cursor. This means you only need to calculate where you want the anchor to be :)

Hmm, doesn't that just mean that we don't have to use `col('.')` etc? The anchor and size we still have to do ourselves :(
> `left`/`right`, `above`/`below`

Would this not be better left up to the implementation to determine where there's space? I.e. to the left if the completion menu is close to the right, and to the right if it's not? Similar to the cursor case, you don't care whether it's above or below the cursor, it should take one or the other depending on where there's enough space.
 
> Probably, unless `cursor_tooltips` and `pum_tooltips` have too much differing internal details.

True, we can always see about that later.

> Good question, especially in combination with the "focusing" via double-tapping. I don't know.
> 
> Maybe best to make that an option, too?

Like the Vim `'moved'` option, I guess.

> Yes; if you put it like that, there's not much to do, though...

Not that that's a bad thing :p

> Yes. (What would be the alternative? If Markup: yes, parsing would be higher-level -- unless you think it might make sense to push this down, in which case the function needs more options.)

I'd keep it at String[] and do the markdown higher up, indeed.

> Yes; possibly including resizing here? (If you move the tooltip to the border, it might have to shrink. Then you'd want to not do this in `set_tooltip_contents`. (All this would be hidden by the top-level API, anyway.)

Ah yeah, you would need to do that too. Probably have a private function `recalc_size` that is called after setting the contents or changing position.
In my mind, it would go: first create popup, then position it, then set the contents (once or more), so I didn't include resizing in positioning, but you're right, we should.

> Yes. I would suggest starting top down: Going through the LSP functions and writing the API calls we _want_ to have, and then implementing them using made-up lower-level calls, and then writing them separately and independently for vim and neovim (probably iterating whenever we hit a snag).

I'll start on a vimdoc for the top level, just to precisely write down what we want each function to do. Can I push it on your branch? Feel free to edit it if think we can improve something.

clason:
> Hmm, doesn't that just mean that we don't have to use `col('.')` etc? The anchor and size we still have to do ourselves :(

True, but it shows the direction neovim is going in, at least for the lower level.

> Would this not be better left up to the implementation to determine where there's space? I.e. to the left if the completion menu is close to the right, and to the right if it's not? Similar to the cursor case, you don't care whether it's above or below the cursor, it should take one or the other depending on where there's enough space.

Agreed.

> Not that that's a bad thing :p

Indeed :)

> Ah yeah, you would need to do that too. Probably have a private function `recalc_size` that is called after setting the contents or changing position.

Yes, that makes sense.

> In my mind, it would go: first create popup, then position it, then set the contents (once or more), so I didn't include resizing in positioning, but you're right, we should.

I wonder if that should be a separate (private) function. But it might need too much state?

> I'll start on a vimdoc for the top level, just to precisely write down what we want each function to do. Can I push it on your branch? Feel free to edit it if think we can improve something.

Sure, that's why I gave you permission!

thomasfaingnaert:
Just thought of something else. For the syntax ranges, we just have filetype at the moment, but it might actually be more useful to have `{'range': {'start': {'row':..., 'col': ...}, 'end': ...}, 'highlight': 'vim highlight group'}`.
If you want syntax highlighting, you use a highlight group that has a `hl-contains`.
But it would allow something like this: preprocess the markdown to strip the *, but remember the position. Then provide that range with a markdown bold highlight group to the API. This would hide the *, similar to conceal, but wouldn't suffer from the problem of having two invisible characters which mess up the sizing. (Yes, I'm still not over that :p)

clason:
I see what you did there ;) (and in your other PR)

Such a function (markdown to highlight group) might be of independent interest -- in fact, it might already exist? (Good starting point would be https://github.com/gabrielelana/vim-markdown)

Next step: tree-sitter integration :) (Yes, there's a LaTeX grammar.)

thomasfaingnaert:
> Such a function (markdown to highlight group) might be of independent interest -- in fact, it might already exist? (Good starting point would be https://github.com/gabrielelana/vim-markdown)

That's probably going to be the hardest problem, if you want to do it correctly and support nesting. For now, a basic implementation would suffice probably. (But that's an interesting link; will definitely check that out)

> Next step: tree-sitter integration :) (Yes, there's a LaTeX grammar.)

Tree-sitter was actually on my TODO list to check out. Have you already tested it?

clason:
> Tree-sitter was actually on my TODO list to check out. Have you already tested it?

No, except in Atom form (briefly, because why switch from vim?), but I've been following neovim's plans to (eventually) integrate it. It looks... non-trivial.

thomasfaingnaert:
Pushed small basic vimdoc, feel free to expand.

clason:
Looks good! Comments, for discussion:

1. API contract should only cover top-level API, others might change as needed. (Meaning vim-lsp can rely on the API stability only for the top level, and should avoid calling intermediate level or below).
2. With this in view, I'd promote `parse_lsp_response` to top-level (and possibly move it to `vim/utils)?
3. If just want syntax highlighting (or none) according to filetype is wanted, passing an empty `[]` for `syn_ranges` is probably easiest? Should be mentioned, then.
4. The `data` fields are `content` and `syn_ranges`, right -- unless there are better names? I'd mention that explicitly, then.

thomasfaingnaert:
>     1. API contract should only cover top-level API, others might change as needed. (Meaning vim-lsp can rely on the API stability only for the top level, and should avoid calling intermediate level or below).

Yes, the API will be strictly hierarchical, I just added it in the documentation to have a quick overview, as a sort of summary of our discussion, I guess. The other layers  should be removed later.

>     2. With this in view, I'd promote `parse_lsp_response` to top-level (and possibly move it to `vim/utils)?

Yeah, I basically regard it as "top-level, but LSP specific", i.e. we would have a function that calls `parse_lsp_response`, and uses that return value for the top level, non-lsp specific API. (Separated it just if we decide to extract this later).

>     3. If just want syntax highlighting (or none) according to filetype is wanted, passing an empty `[]` for `syn_ranges` is probably easiest? Should be mentioned, then.

Sure, I'll add that.

>     4. The `data` fields are `content` and `syn_ranges`, right -- unless there are better names? I'd mention that explicitly, then.

For hover and documentation, you would call `parse_lsp_response` with the data you get from the language server, and pass the return value as `{lines}` and `{syn-ranges}` to the top-level API. For peek definition, you would just read the correct range and not call `parse_lsp_response`. Feel free to suggest better names, though.

jerdna-regeiz:
Nice work =) 
When working out a more high level api for the popup, keep in mind: there are features in both vim and neovim that the other just does not have at all. Like focus the popup or borders. 

positioning at the cursor (as anchor) was already possible when I wrote it. It did not make the code any easier, as the size and position (above/below and left/right) calculation requires the same information (in fact it was some how uglier).

I would not recommend leaving the old preview behind as by far not  everybody has a recent vim or neovim at hand.

Still have to look at the code itself, I liked what I saw at a first glance.

thomasfaingnaert:
@jerdna-regeiz 
I think the code itself is put on hold for a bit :)
I definitely would not want to abandon preview, but it's really different from float, so we decided to ignore it a little bit.
Probably worth revisiting that now, though, before expanding on the API too much.
Feel free to suggest improvements as well, we really need feedback.

@clason Maybe adding @jerdna-regeiz as a collaborator would be good as well?

thomasfaingnaert:
> When working out a more high level api for the popup, keep in mind: there are features in both vim and neovim that the other just does not have at all. Like focus the popup or borders.

That's also something that needs discussing. I assume we still want to keep the doubletap functionality, which is only possible in Neovim at the moment.
For borders, theoretically we can create a second float behind the first to emulate a border, but I think that's overkill; we better wait until neovim upstream handles borders.

clason:
> Yes, the API will be strictly hierarchical, I just added it in the documentation to have a quick overview, as a sort of summary of our discussion, I guess. The other layers  should be removed later.

Can be kept, because more documentation = more better, just with a caveat.

> Yeah, I basically regard it as "top-level, but LSP specific", i.e. we would have a function that calls `parse_lsp_response`, and uses that return value for the top level, non-lsp specific API. (Separated it just if we decide to extract this later).

Makes sense 👍 

> For hover and documentation, you would call `parse_lsp_response` with the data you get from the language server, and pass the return value as `{lines}` and `{syn-ranges}` to the top-level API. For peek definition, you would just read the correct range and not call `parse_lsp_response`. Feel free to suggest better names, though.

That comment was fired from the hip, ignore that. It's fine as is.

@jerdna-regeiz 

> When working out a more high level api for the popup, keep in mind: there are features in both vim and neovim that the other just does not have at all. Like focus the popup or borders.

Good point; these things will go in options, and the backend just ignore options it can't handle (yet). I actually prefer no borders to vim's massive Great Walls, but YMMV. (Nested floats look neat, but are not worth the effort with resizing/positioning).

> positioning at the cursor (as anchor) was already possible when I wrote it. It did not make the code any easier, as the size and position (above/below and left/right) calculation requires the same information (in fact it was some how uglier).

True; the `row`/`col` need to be fudged a bit. Hopefully the high-level API of vim and neovim will converge a bit, and thus it makes sense to try to use their features as much as possible (within reason, of course).

> I would not recommend leaving the old preview behind as by far not everybody has a recent vim or neovim at hand.

I had assumed that people who work with language servers are not upgrade averse, but may be wrong, of course. The problem is that preview only makes sense for signature help and documentation, while hover and peek don't. In particular, signature is a tooltip feature for popup and float, but a documentation-like feature for preview. That makes everything hard (but not impossible) to separate cleanly. But the current proposal should make it easy to add new backends, including preview. (I'd just prefer not to have to worry about that at first.)

And yes, the current code is garbage.

> @clason Maybe adding @jerdna-regeiz as a collaborator would be good as well?

Done!

thomasfaingnaert:
> And yes, the current code is garbage.

Keep in mind you're speaking to the person who wrote it :p

clason:
No, I meant the current code _in this PR_! The current code on master is working fine :)

thomasfaingnaert:
>     4. `options` should be options for the _tooltip_, not the content, presumably. This would be one place to specify `{'close':{'on_move',...}}` -- or other way around `{'on_move':'close'/'follow'}`? Also, `noborder` for vim popup (or `border` if `noborder` is made default)?

That's the plan. Can you add the options to the doc (and any additional ones, if you think of them).

(As an aside; once the academic year resumes next week, I'll probably not have much time anymore. I really hope this gains some traction so I don't have to miss the progress :) )

clason:
> That's the plan. Can you add the options to the doc (and any additional ones, if you think of them).

Will do -- once I come to a conclusion on which way around ;)

> (As an aside; once the academic year resumes next week, I'll probably not have much time anymore. I really hope this gains some traction so I don't have to miss the progress :) )

We still have four more weeks, but these will be _very_ busy weeks for me -- moreso as I've been procrastinating here a bit ;)

thomasfaingnaert:
> We still have four more weeks, but these will be _very_ busy weeks for me -- moreso as I've been procrastinating here a bit ;)

I noticed. Now I'm actually wondering if my professors work on their Vim config during the holidays ;)

clason:
If their definition of a holiday is "a slow week at the office, and a grant proposal to write", possibly -- unless they use Emacs. (Or, more likely, TeXShop.)

clason:
closed in favor of #510 to focus on the API design.

@thomasfaingnaert @jerdna-regeiz @hrsh7th 

(To avoid littering that issue with too many back-and-forth comments, some discussion could be moved to gitter -- a vim-lsp chatroom might be useful.)

thomasfaingnaert:
@clason Haven't used Gitter before. Can I create the room or must the repo owner do that?

clason:
Good question! You could simply try?

On 14 Sep 2019 13:56, Thomas Faingnaert <notifications@github.com> wrote:

@clason<https://github.com/clason> Haven't used Gitter before. Can I create the room or must the repo owner do that?

thomasfaingnaert:
Seems like I can create a community, but if I want to link it to the GitHub repo, I need to own it.
Let's hear @prabirshrestha's thoughts first

clason:
Makes sense, although that would only get the fancy feed and linking to issues/PRs -- for discussion, it doesn't have to be linked. (I'm worried of having 100+ comments on the new PR in the first day already -- I tend to think by writing ;))

thomasfaingnaert:
Also seems that I can't create a community with the name `vim-lsp` (I guess because a GitHub organisation with the same name already exists: https://github.com/vim-lsp)

clason:
Ah. In this case, it would indeed be better for @prabirshrestha to do this (and maybe move the vim-lsp(-related) repo(s) to this organization)?

thomasfaingnaert:
I think it is someone else's, but still `gitter.im/prabirshrestha/vim-lsp` is probably better than `gitter.im/vim-lsp-and-some-random-suffix-i-came-up-with`

thomasfaingnaert:
Let's just create a gitter for your fork for now, might be easier?

clason:
Then it might be possible (and reasonable) to appropriate it; I think github has procedures for that.
https://help.github.com/en/articles/github-username-policy

Otherwise I agree; but I can definitely create a temporary one for this PR. I don't have an organization, but I just named it `clason-vim-lsp`. It'll be deleted later anyway.

prabirshrestha:
Even I find Gitter easier to use. Let me know what name you want. It would be good to have a permanent one.

PR #521, opened 2019-09-25
Add `g:lsp_virtual_text_prefix` option

coditva:
This option allows prepending characters or text to the diagnostics
which are display as virtual text. It allows virtual text to be
differentiated from the code.

Allows me to make it so:
Screenshot 2019-09-25 at 19 28 08


clason:
Yes, that looks much nicer! @thomasfaingnaert Tested, LGTM!

(Arguably, such a symbol (maybe `> ` or similar to be safe with respect to fonts) might even be a good default rather than none.)

thomasfaingnaert:
@coditva @clason Merged. Thanks for implementing/reviewing!

PR #524, opened 2019-09-28
Allow custom server-specific symbol and completion item kind

### LINKED ISSUE #486: Allow per-server customization of symbol and completion kinds
The standard enum from the [LSP specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol) is a poor fit for some languages (such as LaTeX, for which there is https://github.com/latex-lsp/texlab).

It would be nice to be able to override the `s:symbol_kinds` and `s:kind_text_mappings` with a custom enum in the `lsp_setup` server configuration.

clason:
Fixes #486 

Replaces flat `s:kind_mappings` by a dictionary keyed on server name that is passed through the calls. Defaults to the standard mappings that can be overridden by a config key, e.g.,
```viml
'config': {
                \     'symbol_kinds': {
                \            '2': 'sec',
                \            '5': 'thm',
                \            '6': 'float',
                \            '7': 'PhD',
                \            '8': 'label',
                \            '11': 'misc',
                \            '12': 'cmd',
                \            '14': 'math',
                \            '22': 'item',
                \            '23': 'book',
                \            '24': 'artl',
                \            '25': 'part',
                \            '26': 'coll',
                \      }, 
\  }
``` 
Should not break current direct use of `default_get_completion_items` (tests pass, `ncm2-vim-lsp` works), but of course the override doesn't get picked up in this case (you need to pass `server_name` after `item`; I'll make a PR for `ncm2-vim-lsp` once this is merged).

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -110,16 +110,20 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
 endfunction
 
 function! lsp#omni#get_completion_item_kind_text(server, completion_item) abort
-    if !has_key(s:completion_item_kinds, a:server)
-        let l:server_info = lsp#get_server_info(a:server)
-        if has_key (l:server_info, 'config') && has_key(l:server_info['config'], 'completion_item_kinds')
-            let s:completion_item_kinds[a:server] = 
-                        \ extend(s:default_completion_item_kinds, l:server_info['config']['completion_item_kinds'])
-        else
-            let s:completion_item_kinds[a:server] = s:default_completion_item_kinds
+    if empty(a:server) 
+        let l:completion_item_kinds = s:default_completion_item_kinds
+    else
+        if !has_key(s:completion_item_kinds, a:server)
+            let l:server_info = lsp#get_server_info(a:server)
+            if has_key (l:server_info, 'config') && has_key(l:server_info['config'], 'completion_item_kinds')
+                let s:completion_item_kinds[a:server] = 
+                            \ extend(s:default_completion_item_kinds, l:server_info['config']['completion_item_kinds'])
+            else
+                let s:completion_item_kinds[a:server] = s:default_completion_item_kinds
+            endif
         endif
+        let l:completion_item_kinds = s:completion_item_kinds[a:server]
     endif
-    let l:completion_item_kinds = s:completion_item_kinds[a:server]
 
     return has_key(a:completion_item, 'kind') && has_key(l:completion_item_kinds, a:completion_item['kind']) 
                 \ ? l:completion_item_kinds[a:completion_item['kind']] : ''
@@ -195,8 +199,9 @@ function! s:remove_typed_part(word) abort
     return strpart(a:word, l:overlap_length)
 endfunction
 
-function! lsp#omni#default_get_vim_completion_item(item, server_name, ...) abort
-    let l:do_remove_typed_part = get(a:, 1, 0)
+function! lsp#omni#default_get_vim_completion_item(item, ...) abort
+    let l:server_name = get(a:, 1, '')
+    let l:do_remove_typed_part = get(a:, 2, 0)
 
     if g:lsp_insert_text_enabled && has_key(a:item, 'insertText') && !empty(a:item['insertText'])
         if has_key(a:item, 'insertTextFormat') && a:item['insertTextFormat'] != 1
@@ -214,7 +219,7 @@ function! lsp#omni#default_get_vim_completion_item(item, server_name, ...) abort
         let l:word = s:remove_typed_part(l:word)
     endif
 
-    let l:kind = lsp#omni#get_completion_item_kind_text(a:server_name, a:item)
+    let l:kind = lsp#omni#get_completion_item_kind_text(l:server_name, a:item)
 
     let l:completion = {
                 \ 'word': l:word,
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -1,6 +1,6 @@
 " constants {{{
 
-let s:kind_text_mappings = {
+let s:default_completion_item_kinds = {
             \ '1': 'text',
             \ '2': 'method',
             \ '3': 'function',
@@ -27,6 +27,8 @@ let s:kind_text_mappings = {
             \ '24': 'operator',
             \ '25': 'type parameter',
             \ }
+
+let s:completion_item_kinds = {}
 
 let s:completion_status_success = 'success'
 let s:completion_status_failed = 'failed'
@@ -96,7 +98,7 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
         return
     endif
 
-    let l:result = s:get_completion_result(a:data)
+    let l:result = s:get_completion_result(a:server_name,a:data)
     let l:matches = l:result['matches']
 
     if g:lsp_async_completion
@@ -107,8 +109,20 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
     endif
 endfunction
 
-function! lsp#omni#get_kind_text(completion_item) abort
-    return has_key(a:completion_item, 'kind') && has_key(s:kind_text_mappings, a:completion_item['kind']) ? s:kind_text_mappings[a:completion_item['kind']] : ''
+function! lsp#omni#get_completion_item_kind_text(server, completion_item) abort
+    if !has_key(s:completion_item_kinds, a:server)
+        let l:server_info = lsp#get_server_info(a:server)
+        if has_key (l:server_info, 'config') && has_key(l:server_info['config'], 'completion_item_kinds')
+            let s:completion_item_kinds[a:server] = 
+                        \ extend(s:default_completion_item_kinds, l:server_info['config']['completion_item_kinds'])
+        else
+            let s:completion_item_kinds[a:server] = s:default_completion_item_kinds
+        endif
+    endif
+    let l:completion_item_kinds = s:completion_item_kinds[a:server]
+
+    return has_key(a:completion_item, 'kind') && has_key(l:completion_item_kinds, a:completion_item['kind']) 
+                \ ? l:completion_item_kinds[a:completion_item['kind']] : ''
 endfunction
 
 " auxiliary functions {{{
@@ -140,7 +154,7 @@ function! s:send_completion_request(info) abort
                 \ })
 endfunction
 
-function! s:get_completion_result(data) abort
+function! s:get_completion_result(server_name, data) abort
     let l:result = a:data['response']['result']
 
     if type(l:result) == type([])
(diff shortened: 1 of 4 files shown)


autoload/lsp/omni.vim
```diff
@@ -199,7 +213,8 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     if l:do_remove_typed_part
         let l:word = s:remove_typed_part(l:word)
     endif
-    let l:kind = lsp#omni#get_kind_text(a:item)
+
+    let l:kind = lsp#omni#get_completion_item_kind_text(a:server_name,a:item)
```
thomasfaingnaert: This might help ;)
```suggestion
    let l:kind = lsp#omni#get_completion_item_kind_text(a:server, a:item)
```

clason: No, it doesn't -- sorry, the commit was a hip shot from my working (hard coded) state; when I tested it, I used the correct name, of course :)

thomasfaingnaert:
Completion seems to be working fine in 4b6f94e, what was the problem you were having?

clason:
Ah, you're right, if I manually trigger it, it works. So it's actually `ncm2-vim-lsp` that's broken by this change :(

thomasfaingnaert:
I'm guessing it calls `lsp#omni#get_vim_completion_item` without the server parameter?

thomasfaingnaert:
I think you can just change this line to make it work: https://github.com/ncm2/ncm2-vim-lsp/blob/5feab9e3334d5a6a427b19c0ce39c7689d5b6a4d/plugin/ncm2_vim_lsp.vim#L97.

clason:
Exactly. Easy to fix, and backward compatible. I'll make a PR.

clason:
Hmm, having trouble getting the tests to pass. Can you explain to me what the splatted arguments are supposed to do here?

clason:
Hmm. Tests are failing because they're not running a language server, and `lsp#get_server_info` does not handle the case that the passed name is not a registered server. 

I think the correct thing to do would be to add `has_key` checks to that function (and `get_server_capabilities` that print an error message and return an empty dictionary (that can be checked downstream) if the server name is not valid. @prabirshrestha

clason:
@thomasfaingnaert @prabirshrestha I think this is ready to merge now.

autoload/lsp/omni.vim
```diff
@@ -370,7 +390,7 @@ function! s:get_cursor_pos_and_edit_length(text_edit) abort
 endfunction
 
 function! lsp#omni#get_completion_item_kinds() abort
```
thomasfaingnaert: Same here with the `valueSet`.

autoload/lsp/ui/vim/utils.vim
```diff
@@ -176,12 +178,21 @@ function! s:is_file_uri(uri) abort
...
+            let s:symbol_kinds[a:server] = extend(s:default_symbol_kinds, l:server_info['config']['symbol_kinds'])
+        else
+            let s:symbol_kinds[a:server] = s:default_symbol_kinds
+        endif
+    endif
+        let l:symbol_kinds = s:symbol_kinds[a:server]
+    return has_key(l:symbol_kinds, a:kind) ? l:symbol_kinds[a:kind] : 'unknown symbol ' . a:kind
 endfunction
 
 function! lsp#ui#vim#utils#get_symbol_kinds() abort
```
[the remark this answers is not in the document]
clason: I thought about that, but the `valueSet` contains only the numbers, not the text, so it shouldn't make a difference (it could be longer, but that would be against the specification, so the server can't be expected to deal with it?

autoload/lsp/ui/vim/utils.vim
```diff
@@ -176,12 +178,21 @@ function! s:is_file_uri(uri) abort
     return stridx(a:uri, 'file:///') == 0
 endfunction
 
-function! s:get_symbol_text_from_kind(kind) abort
-    return has_key(s:symbol_kinds, a:kind) ? s:symbol_kinds[a:kind] : 'unknown symbol ' . a:kind
+function! s:get_symbol_text_from_kind(server, kind) abort
+    if !has_key(s:symbol_kinds, a:server)
+        let l:server_info = lsp#get_server_info(a:server)
+        if has_key (l:server_info, 'config') && has_key(l:server_info['config'], 'symbol_kinds')
+            let s:symbol_kinds[a:server] = extend(s:default_symbol_kinds, l:server_info['config']['symbol_kinds'])
```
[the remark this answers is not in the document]
clason: Yes, you're right; I'll fix that.

autoload/lsp/ui/vim/utils.vim
```diff
@@ -176,12 +178,21 @@ function! s:is_file_uri(uri) abort
...
+        else
+            let s:symbol_kinds[a:server] = s:default_symbol_kinds
+        endif
+    endif
+        let l:symbol_kinds = s:symbol_kinds[a:server]
```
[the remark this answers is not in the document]
clason: With pleasure :)

autoload/lsp/omni.vim
```diff
@@ -96,7 +98,7 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
         return
     endif
 
-    let l:result = s:get_completion_result(a:data)
+    let l:result = s:get_completion_result(a:server_name,a:data)
```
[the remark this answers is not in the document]
clason: RIghtly so :)

autoload/lsp/omni.vim
clason: See above.

autoload/lsp/omni.vim
```diff
@@ -107,8 +109,24 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
     endif
 endfunction
 
-function! lsp#omni#get_kind_text(completion_item) abort
-    return has_key(a:completion_item, 'kind') && has_key(s:kind_text_mappings, a:completion_item['kind']) ? s:kind_text_mappings[a:completion_item['kind']] : ''
+function! lsp#omni#get_completion_item_kind_text(server, completion_item) abort
```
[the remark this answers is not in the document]
clason: Ah, I didn't think it would be called externally. You're right; I'll name it back and make server optional here as well.

autoload/lsp/ui/vim/utils.vim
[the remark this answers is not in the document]
clason: squashed all the changes by mistake into the doc commit...

autoload/lsp/ui/vim/utils.vim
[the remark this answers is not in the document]
thomasfaingnaert: I was thinking about adding support for additional kinds, but you're right, that's against the specification. Does texlab have additional kinds, or does it just use the predefined ones?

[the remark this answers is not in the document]
clason: It's not illegal to extend the specification. But texlab sticks to the number in the current specification (26 resp. 25). 

It's possible to do this, but requires some refactoring (making the s:symbol_kinds a global variable that is constructed during registration); to quote Bartleby: I'd rather not. (I.e., leave it for a future PR when someone actually requests this for a specific server.)

doc/vim-lsp.txt
```diff
@@ -625,9 +625,18 @@ The vim |dict| containing information about the server.
...
+    initialization. Configuration settings are language-server specific.
+
+    Example:
+	'workspace_config': {'pyls': {'plugins': \
+		{'pydocstyle': {'enabled': v:true}}}}
+
   * config:
     optional vim |dict|
-    Used to pass additional custom config.
+    Used to pass additional custom config. See |vim-lsp-config-server|.
```
thomasfaingnaert: Can you just add the `vim-lsp-config-server` section (or remove the reference)?

autoload/lsp/ui/vim/utils.vim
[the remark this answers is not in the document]
thomasfaingnaert: You should be able to reuse the code in `lsp#omni#get_kind_text`, but I'm fine with leaving it as-is.

doc/vim-lsp.txt
clason: Oops, forgot to remove that. (But I can also add the section -- although it's very close to that location. What would you prefer?)

thomasfaingnaert: Let's just remove it, otherwise we also have to add an entry to the contents.

autoload/lsp/ui/vim/utils.vim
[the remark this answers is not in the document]
clason: Yes, that's what I meant by "refactor" -- there's no new code necessary, but the current code has to be rearranged somewhat (and exposed in the case of `symbolKind`). Let's leave it for later, then.

thomasfaingnaert:
@clason I've just removed some trailing newlines from the doc. Is this ready to be merged?

clason:
That's fine, I'm not too familiar with the vim-help format (and the mandatory conceal doesn't help ;))

Yes, as far as I'm concerned, this is good to go!

thomasfaingnaert:
Vim doesn't care if there is whitespace at the end of the line; I'm just not a big fan of it ;)
LGTM!

autoload/lsp/omni.vim
```diff
@@ -107,8 +109,25 @@ function! s:handle_omnicompletion(server_name, complete_counter, data) abort
...
-    return has_key(a:completion_item, 'kind') && has_key(s:kind_text_mappings, a:completion_item['kind']) ? s:kind_text_mappings[a:completion_item['kind']] : ''
+function! lsp#omni#get_kind_text(completion_item, ...) abort
+    let l:server = get(a:, 1, '')
+    if empty(l:server) " server name 
+        let l:completion_item_kinds = s:default_completion_item_kinds
+    else
+        if !has_key(s:completion_item_kinds, l:server)
+            let l:server_info = lsp#get_server_info(l:server)
+            if has_key (l:server_info, 'config') && has_key(l:server_info['config'], 'completion_item_kinds')
+                let s:completion_item_kinds[l:server] = s:default_completion_item_kinds
```
thomasfaingnaert: @clason 
Just thinking about this (a little late I know): does this make a copy of the dict or just a reference to it?
I think the following line still changes `s:default_completion_item_kinds`?

clason: Yes, you're right! Sorry, vimL newbie.

It's probably best to write this as
```viml
let s:completion_item_kinds[l:server] = extend(copy(s:default_completion_item_kinds), l:server_info['config']['completion_item_kinds']))
```

thomasfaingnaert: That should work. I should've caught this before merging, I guess we're both newbies :smiley: 
Can you send PR?

clason: Done #529 Thanks for catching that!

PR #525, opened 2019-09-29
Use function instead of commands for text_edit.

hrsh7th:
Change text_edit implementation to use `setbufline`, `appendbufline` and `deletebufline` instead of normal-mode commands.

It is mostly passed test-cases.
But one test case is not passed, I think it is test-case side problem.

### DIFF
--- a/autoload/lsp/utils/text_edit.vim
+++ b/autoload/lsp/utils/text_edit.vim
@@ -7,33 +7,6 @@ function! lsp#utils#text_edit#apply_text_edits(uri, text_edits) abort
     " not interfere with the position of later edits, they need to be applied
     " one at the time or put together as a single command.
     "
-    " Example: {"range": {"end": {"character": 45, "line": 5}, "start":
-    " {"character": 45, "line": 5}}, "newText": "\n"}, {"range": {"end":
-    " {"character": 45, "line": 5}, "start": {"character": 45, "line": 5}},
-    " "newText": "import javax.ws.rs.Consumes;"}]}}
-    "
-    " If we apply the \n first we will need adjust the line range of the next
-    " command (so the import will be written on the next line) , but if we
-    " write the import first and then the \n everything will be fine.
-    " If you do not apply a command one at  time, you will need to adjust the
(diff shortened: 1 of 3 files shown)


test/lsp/utils/text_edit.vimspec
```diff
@@ -548,7 +548,7 @@ Describe lsp#utils#text_edit
                     \ }])
 
             let l:buffer_text = s:get_text()
-            Assert Equals(l:buffer_text, ['x', 'y', 'z', 'b', ''])
```
hrsh7th: IMO, The buffer is `['foo', 'b']` and range-end is `line: 1, char: 1`.
So `b` character is in range to replacement.

autoload/lsp/utils/text_edit.vim
```diff
@@ -44,42 +17,67 @@ function! lsp#utils#text_edit#apply_text_edits(uri, text_edits) abort
     " ((0, 4), (0, 5), "") - remove fifth character 'e'
     " ((0, 2), (0, 3), "") - remove third character 'c'
     let l:text_edits = sort(deepcopy(a:text_edits), '<SID>sort_text_edit_desc')
-    let l:i = 0
 
+    let l:path = lsp#utils#uri_to_path(a:uri)
+    let l:bufnr = bufnr(l:path, v:true)
+    if !bufloaded(l:path)
+        call bufload(l:path)
```
hrsh7th: Sorry, CI is failed on this line.
`bufload` seems to be a new api...

hrsh7th: I can't found the way to load buffer with no side-effect.
So I will close this PR later...

thomasfaingnaert: @hrsh7th 
What about storing the current buffer in a variable, running `:edit`, and restoring the current buffer again?

hrsh7th: I trying to the solution, thanks. It works well.

https://github.com/prabirshrestha/vim-lsp/compare/master...hrsh7th:change-text-edit-implementation?expand=1

PR #528, opened 2019-10-02
Add lsp#formatexpr()

clktmr:
This patch adds lsp#formatexpr() which can be used to format ranges in
conjunction with the "gq" command (see :h 'formatexpr'). Because
s:document_format_range() would only format based on the visual
selection markers, a new range parameter was added to pass start_lnum
and end_lnum.


thomasfaingnaert:
Wouldn't it make more sense to use equalprg for this?

Op 2 okt. 2019 21:11 schreef Timur Celik <notifications@github.com>:

This patch adds lsp#formatexpr() which can be used to format ranges in
conjunction with the "gq" command (see :h 'formatexpr'). Because
s:document_format_range() would only format based on the visual
selection markers, a new range parameter was added to pass start_lnum
and end_lnum.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/prabirshrestha/vim-lsp/pull/528

Commit Summary

  *   Add lsp#formatexpr()

File Changes

  *   M autoload/lsp.vim<https://github.com/prabirshrestha/vim-lsp/pull/528/files#diff-0> (6)
  *   M autoload/lsp/ui/vim.vim<https://github.com/prabirshrestha/vim-lsp/pull/528/files#diff-1> (17)
  *   M doc/vim-lsp.txt<https://github.com/prabirshrestha/vim-lsp/pull/528/files#diff-2> (9)

Patch Links:

  *   https://github.com/prabirshrestha/vim-lsp/pull/528.patch
  *   https://github.com/prabirshrestha/vim-lsp/pull/528.diff

clktmr:
I looked up the vim-help again, you are right. Most users probably don't
want to change 'formatexpr' because they use it to format comments. And
while testing I noticed this breaks auto wrapping while inserting text.

However 'equalprg' is supposed to be an external program and
'indentexpr' will probably mess with calculating indent of a newline
after inserting a break.

I mainly wanted to be able to use a motion to format a range, so maybe
it would be best to just map it to another key?

thomasfaingnaert:
I think we should just create a function that implements a LspRangeFormat operator.
The user can then map this to any key he wants, even `=` as a way to "customise `equalprg` indirectly".

clktmr:
I updated the PR to only define a new binding. Configuration is up to
the user.

thomasfaingnaert:
@clktmr Can you fix the linter errors?

plugin/lsp.vim
```diff
@@ -96,6 +96,7 @@ nnoremap <plug>(lsp-peek-type-definition) :<c-u>call lsp#ui#vim#type_definition(
 nnoremap <plug>(lsp-workspace-symbol) :<c-u>call lsp#ui#vim#workspace_symbol()<cr>
 nnoremap <plug>(lsp-document-format) :<c-u>call lsp#ui#vim#document_format()<cr>
 vnoremap <plug>(lsp-document-format) :<Home>silent <End>call lsp#ui#vim#document_range_format()<cr>
+nnoremap <plug>(lsp-document-range-format) :set opfunc=lsp#ui#vim#document_range_format_opfunc<cr>g@
```
thomasfaingnaert: Can we add an `xnoremap` as well so you can `=` from visual mode?

clktmr:
> @clktmr Can you fix the linter errors?

Done. Thanks for reviewing!

thomasfaingnaert:
Thanks, merged. Sorry for the delay.

PR #535, opened 2019-10-09
Hotfix for #455

jackguo380:
Fixed a small bug which was a result of #455

From https://github.com/jackguo380/vim-lsp-cxx-highlight/issues/12#issuecomment-539872380
> Thanks, I investigated the issue and its just a bug on vim-lsp's part. I took a closer look at that commit ([prabirshrestha/vim-lsp#455](https://github.com/prabirshrestha/vim-lsp/pull/455)) it turns out the fix is quite simple:
> 
> A lambda is being passed into `timer_start()` but that function calls the lambda and passes in a timer handle which is not what the lambda expects. The fix is to have `result` come from the outer lambda where `result` is the JSON message.
> 
> ```diff
> diff --git a/autoload/lsp.vim b/autoload/lsp.vim
> index ee6f568daf0a7c56e0e1d2b9f3ed823b5f336119..25d97670cc137719b98877a2fd51cfb2885315f0 100644
> --- a/autoload/lsp.vim
> +++ b/autoload/lsp.vim
> @@ -194,7 +194,7 @@ function! s:on_text_document_did_save() abort
>          " We delay the callback by one loop iteration as calls to ensure_flush
>          " can introduce mmap'd file locks that linger on Windows and collide
>          " with the second lang server call preventing saves (see #455)
> -        call s:ensure_flush(l:buf, l:server_name, {result->timer_start(0, {result->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})})
> +        call s:ensure_flush(l:buf, l:server_name, {result->timer_start(0, {timer->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})})
>      endfor
>  endfunction
>  
> ```

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -194,7 +194,7 @@ function! s:on_text_document_did_save() abort
         " We delay the callback by one loop iteration as calls to ensure_flush
         " can introduce mmap'd file locks that linger on Windows and collide
         " with the second lang server call preventing saves (see #455)
-        call s:ensure_flush(l:buf, l:server_name, {result->timer_start(0, {result->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})})
+        call s:ensure_flush(l:buf, l:server_name, {result->timer_start(0, {timer->s:call_did_save(l:buf, l:server_name, result, function('s:Noop'))})})
     endfor
 endfunction
 


mattn:
Thank you

thenewvu:
@mattn @jackguo380 I think this PR causes #606, as the timer_start() doc said, the only argument of the second lambda in the diff is the timer ID with whatever name. Could someone explain to me what is this change for?

mattn:
The second argument of s:call_did_save must be result not timer. So this PR fix the variable name.

thenewvu:
But look at the diff, it's not what suppose to be right?

mattn:
second argument of timer_start is FuncRef. And the FuncRef take one argument "timer".

```
						*timer_start()* *timer* *timers*
timer_start({time}, {callback} [, {options}])
		Create a timer and return the timer ID.

		{time} is the waiting time in milliseconds. This is the
		minimum time before invoking the callback.  When the system is
		busy or Vim is not waiting for input the time will be longer.

		{callback} is the function to call.  It can be the name of a
		function or a |Funcref|.  It is called with one argument, which
		is the timer ID.  The callback is only invoked when Vim is
		waiting for input.

		{options} is a dictionary.  Supported entries:
		   "repeat"	Number of times to repeat calling the
				callback.  -1 means forever.  When not present
				the callback will be called once.
				If the timer causes an error three times in a
				row the repeat is cancelled.  This avoids that
				Vim becomes unusable because of all the error
				messages.

		Example: >
			func MyHandler(timer)
			  echo 'Handler called'
			endfunc
			let timer = timer_start(500, 'MyHandler',
				\ {'repeat': 3})
<		This will invoke MyHandler() three times at 500 msec
		intervals.

		Can also be used as a |method|: >
			GetMsec()->timer_start(callback)

<		Not available in the |sandbox|.
		{only available when compiled with the |+timers| feature}
```

thenewvu:
Yes, it's a lambda in this case right? the argument of this lamda will be a timer ID as the doc said "It is called with one argument, which is the timer ID.". So why does changing its name make any sense? Please correct me, I'm not familar with vimscript, just trying to figure out what causes #606.

mattn:
Original (broken) code make shadowing the variable `result`.

image

thenewvu:
Oh dear, sorry for taking your time, my fault.

mattn:
NP~ :smile:

PR #540, opened 2019-10-19
add line num args to prop_remove

studiome:
Thank you for creating greate plugins.
I use Vim 8.1.0875 at Debian Buster. 
"Not enough arguments for function: prop_remove" error occurs when I use this plugins.
Newer version of Vim solved this problems but Buster's package have a litter older.
I added line number args to the function.
Dose this change work?

### DIFF
--- a/autoload/lsp/ui/vim/folding.vim
+++ b/autoload/lsp/ui/vim/folding.vim
@@ -31,7 +31,7 @@ function! s:set_textprops(buf) abort
     silent! call prop_type_add(s:textprop_name, {'bufnr': a:buf})
 
     " First, clear all markers from the previous run
-    call prop_remove({'type': s:textprop_name, 'bufnr': a:buf})
+    call prop_remove({'type': s:textprop_name, 'bufnr': a:buf}, 1, line('$'))
 
     " Add markers to each line
     let l:i = 1


autoload/lsp/ui/vim/folding.vim
```diff
@@ -31,7 +31,7 @@ function! s:set_textprops(buf) abort
     silent! call prop_type_add(s:textprop_name, {'bufnr': a:buf})
 
     " First, clear all markers from the previous run
-    call prop_remove({'type': s:textprop_name, 'bufnr': a:buf})
+    call prop_remove({'type': s:textprop_name, 'bufnr': a:buf}, 1, line('$'))
```
thomasfaingnaert: @studiome Just wondering, `line('$')` returns the number of lines in the current buffer, however, shouldn't we make sure to get the number of lines in the buffer `a:buf` instead? I remember some problems with not specifying `a:buf` when I implemented folding.

studiome: Using a:buf instead of ‘$’ occurs error “Invalid Range”. Umm...

thomasfaingnaert: Yeah, `line` takes a line expr and a window id (optionally), so you still need the `'$'`. Unfortunately, it doesn't take a buffer id, and I don't know a function that does off the top of my head, so I guess we can leave it until it causes problems.

PR #549, opened 2019-10-25
Open Neovim floating windows with "minimal" style

dradtke:
I've been using `vim-lsp` a fair amount recently, and I noticed that a lot of completion popups didn't seem to set the width appropriately:

Screen Shot 2019-10-25 at 9 54 13 AM

After digging in to this, I realized that the little gutter on the left-hand side of the popup was probably being counted by Neovim as part of the window's width, while `vim-lsp` was setting the width to the length of the longest line exactly.

To fix this, I discovered the `style` option for configuring popup windows:

```
 style : Configure the appearance of the window.
Currently only takes one non-empty value:
• "minimal" Nvim will display the window with
  many UI options disabled. This is useful
  when displaying a temporary float where the
  text should not be edited. Disables
  'number', 'relativenumber', 'cursorline',
  'cursorcolumn', 'foldcolumn', 'spell' and
  'list' options. 'signcolumn' is changed to
  auto . The end-of-buffer region is hidden
  by setting eob flag of 'fillchars' to a
  space char, and clearing the EndOfBuffer
  region in 'winhighlight'.
```

Here's what it looks like:

Screen Shot 2019-10-25 at 9 55 33 AM

Note that the first line fits entirely in the window, rather than wrapping the final two characters to a new line.

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -87,6 +87,7 @@ function! s:get_float_positioning(height, width) abort
       endif
     endif
     let l:col = col('.')
+    let l:style = 'minimal'
     " Positioning is not window but screen relative
     let l:opts = {
           \ 'relative': 'win',
@@ -94,6 +95,7 @@ function! s:get_float_positioning(height, width) abort
           \ 'col': l:col,
           \ 'width': l:width,
           \ 'height': l:height,
+          \ 'style': l:style,
           \ }
     return l:opts
 endfunction


clason:
Yes, that looks like a reasonable change (and is in fact the intended style for such popups).

There's a plan to rework the floats, making better use of the neovim api when it is available, in #510; feel free to add your thoughts there. One thing that is missing currently is to account for soft-wrapping: if the location is at a wrapped line (after the softwrap), the popup placement is off (right-aligned at window for hover, too low for completion).

(@thomasfaingnaert I tested this, and it seems fine to merge.)

thomasfaingnaert:
@clason @dradtke Is this supported in all Neovim versions with a floating window, or do we need to add a version check for this?

clason:
@thomasfaingnaert Yes, the first version with floating windows (0.4.0) already supports the minimal style (https://github.com/neovim/neovim/commit/ef3e32d57ee0ebe3f425998216d8bf61ffdc28bb).

thomasfaingnaert:
Merged, thanks.

PR #557, opened 2019-11-13
Fix canceling visual-mode when folding enabled.

### LINKED ISSUE #556: Unexpectedly canceling visual mode when enable folding feature.
I found the reason.
The reason is below code.
https://github.com/prabirshrestha/vim-lsp/blob/efff12608f334edcc6878ce9a790b136c7fa92c6/autoload/lsp/ui/vim/folding.vim#L189.

It looks like the `windo` canceling visual-mode.

I tried to solve it, but I can't understood the code meanings.
https://github.com/prabirshrestha/vim-lsp/blob/efff12608f334edcc6878ce9a790b136c7fa92c6/autoload/lsp/ui/vim/folding.vim#L190
This line says `foldmethod is 'expr' so assign 'expr' to it`.
Should this fix too?

hrsh7th:
This PR aims to fix #556.

If user in visual-mode, restore the visual-range when update folding.
This approach enables updating folding in visual-mode, I think.

### DIFF
--- a/autoload/lsp/ui/vim/folding.vim
+++ b/autoload/lsp/ui/vim/folding.vim
@@ -178,15 +178,13 @@ function! s:handle_fold_request(server, data) abort
     endif
     let s:folding_ranges[a:server][l:bufnr] = l:result
 
-    " Don't do the 'windo' in Insert mode, it puts Vim back in Normal mode.
-    if mode()[0] ==# 'i'
-        return
-    endif
-
     " Set 'foldmethod' back to 'expr', which forces a re-evaluation of
     " 'foldexpr'. Only do this if the user hasn't changed 'foldmethod',
     " and this is the correct buffer.
-    let l:current_window = winnr()
-    windo if &l:foldmethod ==# 'expr' && bufnr('%') == l:bufnr | let &l:foldmethod = 'expr' | endif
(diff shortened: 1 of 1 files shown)


autoload/lsp/ui/vim/folding.vim
```diff
@@ -187,6 +187,12 @@ function! s:handle_fold_request(server, data) abort
     " 'foldexpr'. Only do this if the user hasn't changed 'foldmethod',
     " and this is the correct buffer.
     let l:current_window = winnr()
+    let l:oldmode = mode()
```
mattn: Thanks your fix.

s/oldmode/old_mode/

hrsh7th: Thanks for your review.
I've fixed and force pushed.

hrsh7th:
I will change the approch to use `getwinvar/setwinvar`.

hrsh7th:
I updated the PR.
I think it is ok to merge.
Please check this.

mattn:
Anyone know neovim have `win_findbuf` ?

hrsh7th:
Yes. I use neovim every day.
neovim has `win_findbuf` api.

mattn:
Thank you

PR #578, opened 2019-11-24
Handle error if the user hits escape before complete() is done

Gavinok:
when using omni complete for lsp completion the user could leave insert mode before the the completion menu is ready. this causes the error `E785: complete() can only be used in Insert mode`.

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -141,8 +141,6 @@ endfunction
 function! s:display_completions(timer) abort
     if mode() is# 'i'
 		call complete(s:start_pos + 1, s:completion['matches'])
-    else
-        return 1
     endif
 endfunction
 
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -140,9 +140,9 @@ endfunction
 
 function! s:display_completions(timer) abort
 	try
-       call complete(s:start_pos + 1, s:completion['matches'])
-	catch /^Vim(call):E785: complete() can only be used in Insert mode/
-	endtry
+        call complete(s:start_pos + 1, s:completion['matches'])
+    catch /^Vim(call):E785: complete() can only be used in Insert mode/
+    endtry
 endfunction
 
 function! s:handle_omnicompletion(server_name, complete_counter, data) abort
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -139,7 +139,10 @@ function! s:contains_filter(item, last_typed_word) abort
 endfunction
 
 function! s:display_completions(timer) abort
-    call complete(s:start_pos + 1, s:completion['matches'])
+	try
+       call complete(s:start_pos + 1, s:completion['matches'])
+	catch /^Vim(call):E785: complete() can only be used in Insert mode/
+	endtry
 endfunction
 
 function! s:handle_omnicompletion(server_name, complete_counter, data) abort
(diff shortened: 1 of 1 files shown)


mattn:
Could you please explain how to reproduce this issue?

Gavinok:
omniError

Basically all I did was call omnicomplete then hit escape. This causes the error seen above.

mattn:
Hmm, can't reproduce this.

Gavinok:
Interesting I tried it with the minimal config and the python language server, using the omnicomplete function. Here is a link to a gist containing the entire [vimrc](https://gist.github.com/Gavinok/559910ac258fb342962f9a9349d72943). The issue still persisted with these settings. Something that may be contributing the issue is that my computer is relatively old(thinkpad x200 with a core 2 duo). So that may be contributing to why vim can not process all the completions in time.

prabirshrestha:
I think it is due to the event loop so might be difficult to repro or if using slower sources/maching.

https://github.com/prabirshrestha/vim-lsp/blob/7049ddeb40d8abcf5ff3e1303be50c3b0331e18d/autoload/lsp/omni.vim#L91

Either way we should check for insert mode instead of use try catch. In asyncomplete I explicitly ignore if we are not in insert mode using `if mode() isnot# 'i'`. Can you update the PR based on this check and see if that solves the issue?

Gavinok:
Worked like a charm, I implemented it similar to how you have it in asyncomplete for continuities sake.

autoload/lsp/omni.vim
```diff
@@ -88,13 +88,22 @@ function! lsp#omni#complete(findstart, base) abort
 
             let s:completion['status'] = ''
 
+    		if s:should_skip() | return | endif
```
prabirshrestha: Can we inline instead of creating a new function called `s:should_skip()`. 

Also I think you need to return some value instead of just using return.

prabirshrestha: Actually we should move this check in `s:display_completions()` and don't call complete if not in insert mode and remove the try..catch.

Gavinok:
@prabirshrestha as requested I removed the function and moved checking to display_completions()

autoload/lsp/omni.vim
```diff
@@ -148,10 +139,11 @@ function! s:contains_filter(item, last_typed_word) abort
 endfunction
 
 function! s:display_completions(timer) abort
-	try
-        call complete(s:start_pos + 1, s:completion['matches'])
-    catch /^Vim(call):E785: complete() can only be used in Insert mode/
-    endtry
+    if mode() is# 'i'
+		call complete(s:start_pos + 1, s:completion['matches'])
+    else
```
prabirshrestha: else block not required.

prabirshrestha:
@Gavinok added one small comment

autoload/lsp/omni.vim
Gavinok: removed it

autoload/lsp/omni.vim
```diff
@@ -139,7 +139,9 @@ function! s:contains_filter(item, last_typed_word) abort
 endfunction
 
 function! s:display_completions(timer) abort
-    call complete(s:start_pos + 1, s:completion['matches'])
+    if mode() is# 'i'
+		call complete(s:start_pos + 1, s:completion['matches'])
```
prabirshrestha: should use 4 spaces and not tabs.

prabirshrestha:
Merged. Thanks.

PR #600, opened 2019-12-15
Fix error on s:set_textprops()

thinca:
`bufwinid()` works on current tab page only, so when user changes the
current tab page, bufwinid() returns -1.


mattn:
LGTM

mattn:
Thank you

PR #605, opened 2019-12-18
Unlet exited in server

mattn:
https://github.com/prabirshrestha/vim-lsp/pull/604 stop server. But when the server start again, exited is not unset.


tyru:
Thanks!
Confirmed that this PR fixes `:LspStatus` issue in #596, and also `:LspStopServer` completion is working!

mattn:
Thank you

PR #661, opened 2020-01-06
Add support for window/showMessageRequest

dradtke:
I was adding the [Scala language server](https://scalameta.org/metals/) to a project, and it wasn't working well because the server prompts the user before importing the SBT configuration. Without `window/logMessage` support, the SBT configuration never gets imported.

This is a very simple implementation that uses `input()` to prompt the user for input based on the actions provided. A more sophisticated one may be possible with popup windows, but this was the fastest way to build it, and also is supported in any version of Vim or Neovim.

As an example, the SBT prompt looks like this:

```
New sbt workspace detected, would you like to import the build?

0. <cancel>
1. Import build
2. Not now
3. Don't show again

> 
```

The first option is always added, and simulates clicking the X on the dialog within a more GUI-ful editor. It expects the user to type a number from `0` to `3`; any other input causes the prompt to be issued again.

No tests have been added yet, but they can be added if needed. I primarily wanted to get unblocked on my Scala project, and to share the solution in case others will find it useful.


autoload/lsp.vim
```diff
@@ -695,6 +695,8 @@ function! s:on_notification(server_name, id, data, event) abort
                 call lsp#ui#vim#diagnostics#handle_text_document_publish_diagnostics(a:server_name, a:data)
             elseif l:response['method'] ==# 'textDocument/semanticHighlighting'
                 call lsp#ui#vim#semantic#handle_semantic(a:server_name, a:data)
+            elseif l:response['method'] ==# 'window/logMessage'
```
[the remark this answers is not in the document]
dradtke: Yeah, I'm fine not including the `logMessage` change. I think it would be nice to be able to catch these somehow, but I agree that using `echo` is probably too intrusive. I included it primarily to help debug `showMessageRequest`, but I can remove it from this PR.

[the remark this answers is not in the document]
dradtke: I just updated the PR to make `lsp#ui#vim#window#log_message` a noop, so that it can be updated in the future to save that message to a logfile, or some other option that's less intrusive.

autoload/lsp/ui/vim/window.vim
```diff
@@ -0,0 +1,52 @@
...
+function! s:echo_end()
+  echohl None
+endfunction
+
+function! lsp#ui#vim#window#log_message(type, message) abort
+  " TODO: find a way to handle log messages. Most editors I think use this to
+  " update the status bar, but that doesn't work quite as well in Vim.
+endfunction
+
+function! lsp#ui#vim#window#show_message_request(type, message, actions) abort
```
hrsh7th: You can use `inputlist` instead.

dradtke:
I just updated the PR to use `inputlist()` instead of `input()`. I also added a placeholder for `window/showMessage`, which ideally would use a popup window like other editors, but the preview method doesn't yet support showing a docked window.

autoload/lsp.vim
```diff
@@ -700,10 +700,16 @@ function! s:on_notification(server_name, id, data, event) abort
...
         if has_key(l:response, 'method')
-            if g:lsp_diagnostics_enabled && l:response['method'] ==# 'textDocument/publishDiagnostics'
+            let l:method = l:response['method']
+            if g:lsp_diagnostics_enabled && l:method ==# 'textDocument/publishDiagnostics'
                 call lsp#ui#vim#diagnostics#handle_text_document_publish_diagnostics(a:server_name, a:data)
-            elseif l:response['method'] ==# 'textDocument/semanticHighlighting'
+            elseif l:method ==# 'textDocument/semanticHighlighting'
                 call lsp#ui#vim#semantic#handle_semantic(a:server_name, a:data)
+            elseif l:method ==# 'window/logMessage'
+                call lsp#ui#vim#window#log_message(str2nr(l:response['params']['type']), l:response['params']['message'])
```
hrsh7th: We should remove or disable this line?
I think, `window/logMessage` results is noisy to show.

hrsh7th: I try to use this with gopls.
A lot of messages shown.

prabirshrestha: Can we remove `window/logMessage`. It is too much noise even for `lsp#log()`

prabirshrestha:
@dradtke add a comment. also do fix the lint issues about adding `abort` in methods

autoload/lsp.vim
dradtke: It can be super noisy, but I still think `window/logMessage` information needs to be available for people to opt-in to, since some language servers use it for some fairly important information. For example, the Scala language server uses these messages to tell the user that the sbt import has started; without them, it just looks like Vim is sitting there doing nothing.

As an alternative approach that should be less noisy, I updated the implementation to, instead of echoing incoming log messages, set them to a couple global variables: `g:lsp_last_log_message_<type|text>`. The idea is that most users will not see them, but you can opt in to them by including one or both of those variables in your status line, i.e.

```vimscript
set statusline+=%{'['.g:lsp_last_log_message_type.']\ '.g:lsp_last_log_message_text}
```

This way the information does not get lost for those who need it, but it doesn't clog up the logs or the Vim command line. Other editors like VSCode put this information in the status bar, so this is consistent with the experience in other editors as well.

dradtke:
I believe the linting issues are fixed, but I installed `vint`, ran it on my branch, and received ~100 linting errors, so I'm sure I'm not running it correctly.

hrsh7th:
I think it can be merged.
If no more review comments adding, I will merge it.

autoload/lsp.vim
```diff
@@ -700,10 +712,18 @@ function! s:on_notification(server_name, id, data, event) abort
...
     if lsp#client#is_server_instantiated_notification(a:data)
         if has_key(l:response, 'method')
-            if g:lsp_diagnostics_enabled && l:response['method'] ==# 'textDocument/publishDiagnostics'
+            let l:method = l:response['method']
+            if g:lsp_diagnostics_enabled && l:method ==# 'textDocument/publishDiagnostics'
                 call lsp#ui#vim#diagnostics#handle_text_document_publish_diagnostics(a:server_name, a:data)
-            elseif l:response['method'] ==# 'textDocument/semanticHighlighting'
+            elseif l:method ==# 'textDocument/semanticHighlighting'
                 call lsp#ui#vim#semantic#handle_semantic(a:server_name, a:data)
+            elseif l:method ==# 'window/logMessage'
```
prabirshrestha: Do we still need log message if we do nothing?

mattn: Any update?

autoload/lsp.vim
```diff
@@ -766,6 +786,9 @@ function! s:on_request(server_name, id, request) abort
     elseif a:request['method'] ==# 'workspace/configuration'
         let l:response_items = map(a:request['params']['items'], { key, val -> lsp#utils#workspace_config#get_value(a:server_name, val) })
         call s:send_response(a:server_name, { 'id': a:request['id'], 'result': l:response_items })
+    elseif a:request['method'] ==# 'window/showMessageRequest'
+        let l:response = lsp#ui#vim#window#show_message_request(str2nr(a:request['params']['type']), a:request['params']['message'], get(a:request['params'], 'actions', []))
```
prabirshrestha, line 790: why use `str2nr` instead of just leaving it as it is?

ckipp01:
> I was adding the Scala language server to a project, and it wasn't working well because the server prompts the user before importing the SBT configuration. Without window/logMessage support, the SBT configuration never gets imported.

Just to clarify this a bit,  Metals uses `window/showMessageRequest`s in quite a few different scenarios. There are also a couple scenarios that without it, it'd be _very_ difficult to even use Metals. So this is a great feature to be added to help support the Metals use-case. The `window/logMessage` part of this is actually completely separate and more related to #871. Metals makes use of both `window/logMessage` and `window/showMessage` request, unless `window/showMessage` isn't supported, then we pipe everything through `window/logMessage`. So it's important for one of these to be supported for the user to see information from the server, but altogether a separate issue than `window/showMessageRequest`.

prabirshrestha:
In the meantime please give https://github.com/prabirshrestha/vim-lsp/pull/919 a try and will close this PR for now.It doesn't support window/logMessage or window/showMessage but is easy to add. We could most likely add this under a flag. In the mean time you can always intercept manually by adding the following line of code in vimscript.Can folks try this PR instead where I added support for `window/showMessageRequest` it does use newer internal vim apis.

 In the mean time you can always intercept manually by adding the following line of code in vimscript. 
```vim
call lsp#callbag#pipe(
    \ lsp#stream(),
    \ lsp#callbag#filter({x->has_key(x, 'response') && has_key(x['response'], 'method') && x['response']['method'] ==# 'window/logMessage'}),
    \ lsp#callbag#forEach({x->lsp#utils#echo_with_truncation(x['response']['params']['message'])})
    \ )
``` 

We could add a flag for both `logMessage` and `showMessage`. I might be skeptical about default true since it could be spamming users. Also might make sense to have this flag configurable per server instead of just global.

prabirshrestha:
`window/showMessageRequest` is in master. Will work on `window/showMessage` separately. But if anyone wants to take a stab at it feel free.

prabirshrestha:
`$/progress` which is similar to `window/showMessage` is now supported. https://github.com/prabirshrestha/vim-lsp/pull/970 (demo in the PR).

PR #677, opened 2020-01-20
Stop toggling lightline on setting contents of popup

itchyny:
I'm not sure which situation these lines are required, but the code was introduced on June 26, 2019 (#395), and lightline fixes for popup on July 15, 2019 (https://github.com/itchyny/lightline.vim/pull/371) so I believe these aren't required anymore.
Ping to @jerdna-regeiz.

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -149,21 +149,10 @@ function! s:setcontent(lines, ft) abort
     if s:use_vim_popup
         " vim popup
         call setbufline(winbufnr(s:winid), 1, a:lines)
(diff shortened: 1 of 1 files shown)


itchyny:
The problem is that when some exception was thrown while setting the filetype, lightline can't be enabled (https://github.com/itchyny/lightline.vim/issues/409#issuecomment-576093444). One solution is to wrap the `setlocal filetype` by try catch, and another solution is to stop toggling lightline.

jerdna-regeiz:
Yes of course, if lightline now is fixed now, just remove the code! (As it already was mentioned in the code comment :) )

mattn:
Thanks all!

PR #683, opened 2020-01-22
Add an option to use customized function to open location.

ipod825:
This is an update of #517. I don't quite understand @prabirshrestha's comment. Particularly, do I still need to pass a dictionary to the customized location_open function? If so, what should the dictionary contain?

### DIFF
--- a/autoload/lsp/utils/location.vim
+++ b/autoload/lsp/utils/location.vim
@@ -1,7 +1,12 @@
 function! s:open_location(path, line, col) abort
     normal! m'
     let l:buffer = bufnr(a:path)
-    if &modified && !&hidden
+    if exists('g:lsp_custom_open_location')
+        call g:lsp_custom_open_location[0]({
+                    \'path':fnameescape(a:path),
+                    \'line':a:line,
+                    \'col':a:col})
+    elseif &modified && !&hidden
         let l:cmd = l:buffer !=# -1 ? 'sb ' . l:buffer : 'split ' . fnameescape(a:path)
     else
         let l:cmd = l:buffer !=# -1 ? 'b ' . l:buffer : 'edit ' . fnameescape(a:path)
--- a/doc/vim-lsp.txt
+++ b/doc/vim-lsp.txt
@@ -603,6 +603,25 @@ g:lsp_completion_resolve_timeout         *g:lsp_completion_resolve_timeout*
     The `completionItem/resolve` request's timeout value.
     If your vim freeze at `CompleteDone`, you can set this value to 0.
 
+g:lsp_custom_open_location				*g:lsp_custom_open_location*
+    Type: |String|
+    Default: `""`
+
+    Name of the customizable function to jump to a location (e.g. in use of
+    |LspDefinition|, |LspImplementation|, etc.). The function takes a
+    dictionary as input whose keys include "path", "line", "col" that are the
+    target path, line # and column #. Note, however, vim-lsp handles line #
+    and column # automatically. Users only need to open the file with their
+    favorite command as shown in the following exaple.
+
+    Example:
+	function! s:LspGoTo(args)
+	    let l:cmd = 'tabedit '.a:args["path"]
+	    execute l:cmd
+	endfunction
+>
+	let g:lsp_custom_open_location = [function('s:LspGoTo')]
+
 ==============================================================================
 FUNCTIONS                                               *vim-lsp-functions*
 
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -42,6 +42,7 @@ let g:lsp_ignorecase = get(g:, 'lsp_ignorecase', &ignorecase)
 let g:lsp_semantic_enabled = get(g:, 'lsp_semantic_enabled', 0)
 let g:lsp_text_document_did_save_delay = get(g:, 'lsp_text_document_did_save_delay', -1)
 let g:lsp_completion_resolve_timeout = get(g:, 'lsp_completion_resolve_timeout', 200)
+let g:lsp_custom_open_location = get(g:, 'lsp_custom_open_location', [])
 
 let g:lsp_get_vim_completion_item = get(g:, 'lsp_get_vim_completion_item', [function('lsp#omni#default_get_vim_completion_item')])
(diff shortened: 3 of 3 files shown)


ipod825:
Updated with dictionary argument.

autoload/lsp/utils/location.vim
```diff
@@ -1,7 +1,12 @@
 function! s:open_location(path, line, col) abort
     normal! m'
     let l:buffer = bufnr(a:path)
-    if &modified && !&hidden
+    if exists('g:lsp_custom_open_location_fn')
+        let l:cmd = 'call '. g:lsp_custom_open_location_fn.'({
+                    \"path":"'.fnameescape(a:path).'",
+                    \"line":"'.a:line.'",
+                    \"col":"'.a:col.'"})'
+    elseif &modified && !&hidden
```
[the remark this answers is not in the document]
ipod825: I don't know any other way than using `execute` to run `call` with variable arguments. Please let me know if you know how to do this.

Reference: https://stackoverflow.com/questions/29662064/how-to-pass-variables-as-arguments-of-a-function-in-a-vim-script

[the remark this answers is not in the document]
hrsh7th: User's setting example.
```vim
let g:lsp_open_location = [function('s:custom_open_location')]

function s:custom_open_location(context) abort
  execute printf('%s %s', a:context.command, a:context.filename)
  if has_key(a:context, 'lnum')
    call cursor([a:context.lnum, get(a:context, 'col', col('.'))])
  endif
endfunction
```

vim-lsp's handling example.
```vim
call g:lsp_open_location[0]({ ... })
```

This implementation already uses to support custom snippet plugin in vim-lsp.
Could you try it?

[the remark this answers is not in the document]
ipod825: It works. But what's the point of using an array to store the functions?

Besides, do you know how to config async.vim to print error message when the async job fails?
It's so depressing to debug vim-lsp when the client code has bug as no error message will raise.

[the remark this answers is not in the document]
hrsh7th: The point of using an array is vim's naming rule.
Vim needs upper character in first of the variable name fo function reference variable.

I'm using `let g:lsp_log_file = '/tmp/lsp.log'` to debugging vim-lsp.
If you want to show it to echo area, you can override `lsp#log` function (but it is bit noisy. so override should be carefully).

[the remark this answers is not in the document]
ipod825: Thanks! Updated as requested.

hrsh7th:
Could you solve CI errors?

ipod825:
Done.

hrsh7th:
Ah, Sorry, Could you solve inline comments from GitHub actions?

ipod825:
Not sure what you mean, but I resolved the conversation. Please inform me if any other work is needed.

autoload/lsp/utils/location.vim
```diff
@@ -1,7 +1,12 @@
 function! s:open_location(path, line, col) abort
     normal! m'
     let l:buffer = bufnr(a:path)
-    if &modified && !&hidden
+    if exists('g:lsp_custom_open_location_fn')
+        call g:lsp_custom_open_location_fn[0]({
+                    \"path":fnameescape(a:path),
```
hrsh7th: Prefer single quote instead of double quote.

ipod825: Done.

doc/vim-lsp.txt
```diff
@@ -603,6 +603,25 @@ g:lsp_completion_resolve_timeout         *g:lsp_completion_resolve_timeout*
     The `completionItem/resolve` request's timeout value.
     If your vim freeze at `CompleteDone`, you can set this value to 0.
 
+g:lsp_custom_open_location_fn				*g:lsp_custom_open_location_fn*
```
hrsh7th: I don't know the vim-lsp's option naming convention.
But I think `_fn` suffix is not needed here because another option has no suffix.

And Type is `List` now. Could you update docs?

hrsh7th: And sorry, please add default definition in `plugin/lsp.vim`.

ipod825: All done.

autoload/lsp/utils/location.vim
```diff
@@ -1,7 +1,12 @@
 function! s:open_location(path, line, col) abort
     normal! m'
     let l:buffer = bufnr(a:path)
-    if &modified && !&hidden
+    if exists('g:lsp_custom_open_location')
```
hrsh7th: Currently, `g:lsp_custom_open_location` is always defined in `plugin/lsp.vim`.
So this line should be `exists(...) && len()`.

Could you solve this problem?

doc/vim-lsp.txt
```diff
@@ -603,6 +603,25 @@ g:lsp_completion_resolve_timeout         *g:lsp_completion_resolve_timeout*
     The `completionItem/resolve` request's timeout value.
     If your vim freeze at `CompleteDone`, you can set this value to 0.
 
+g:lsp_custom_open_location				*g:lsp_custom_open_location*
+    Type: |String|
```
hrsh7th: Please change to `|List|`.

hrsh7th: `Default: []`

plugin/lsp.vim
```diff
@@ -42,6 +42,7 @@ let g:lsp_ignorecase = get(g:, 'lsp_ignorecase', &ignorecase)
 let g:lsp_semantic_enabled = get(g:, 'lsp_semantic_enabled', 0)
 let g:lsp_text_document_did_save_delay = get(g:, 'lsp_text_document_did_save_delay', -1)
 let g:lsp_completion_resolve_timeout = get(g:, 'lsp_completion_resolve_timeout', 200)
+let g:lsp_custom_open_location = get(g:, 'lsp_custom_open_location', '')
```
hrsh7th: Please change to `[]` to default value.

ipod825: done.

prabirshrestha:
https://github.com/prabirshrestha/vim-lsp/pull/776 introduces mods which aligns with how other vim command works.

Example of usage:

`:leftabove LspDefinition` Splits horizontally and opens a window to the above.
`:rightbelow vertical LspDefinition` Splits vertically and opens a window to the right.

PR #693, opened 2020-01-29
Add sketch of initial hooks implementation

dradtke:
This PR supersedes https://github.com/prabirshrestha/vim-lsp/pull/674, and represents an initial attempt at adding hooks to `vim-lsp`. The idea behind it is that each hook is identified by four parameters:

1. `when`: must be either `before` or `after`, and indicates when the hook fires relative to the event. `before` hooks can block execution by returning a truthy value from the handler; the return value of `after` hooks is ignored.
2. `server_name`: name of the language server the hook should apply to.
3. `event`: must be one of `request`, `response`, or `notification`, and indicates the event that will cause the hook to fire. This corresponds with the [message types](https://microsoft.github.io/language-server-protocol/specifications/specification-current/) defined in the spec.
4. `method`: the method to invoke the hook for.

Currently, these values are concatenated into a single string so that they can serve as a dictionary key. The value is a function that accepts a single parameter, `data`, which contains the necessary event data.

There are still some things to add (notably parameter validation in the `register_hook()` function), and likely some tweaking and polish to ensure it works well as a solution, but I believe it represents a good general solution to the problem of integrating language-specific functionality.

## An Example

Using this, I was able to replicate my fix for `java.apply.workspaceEdit` with the following:

```viml
function! s:handle_client_side_command(data) abort
  let l:method = a:data['method']
  if l:method != 'workspace/executeCommand'
    return 0
  endif

  let l:command = a:data['params']['command']
  let l:arguments = get(a:data['params'], 'arguments', [])

  if l:command ==# 'java.apply.workspaceEdit'
    for l:argument in l:arguments
      call lsp#utils#workspace_edit#apply_workspace_edit(l:argument)
    endfor
    return 1
  endif

  return 0
endfunction

call lsp#register_hook('before', 'java-language-server', 'request', 'workspace/executeCommand', function('s:handle_client_side_command'))
```

This successfully intercepts the `workspace/executeCommand`, preventing it from being sent to the server and instead implementing the command directly in Vim using the existing workspace edit function.


dradtke:
@prabirshrestha Any thoughts on this?

PR #696, opened 2020-01-30
Cache signature_help_trigger_chars

mattn:
s:on_text_changed_after in autoload/lsp/ui/vim/signature_help.vim is too heavy eventhough it is called by every key typing.


hrsh7th:
I'm sorry. It's my bad, thanks!

PR #698, opened 2020-02-01
Fix clear completed text

hrsh7th:


hrsh7th:
I have the completion problem with julia-language-server.
This patch aims to solve the problem, but it couldn't...

PR #699, opened 2020-02-02
Register CursorMoved event only when required

itchyny:
I'm wondering we can improve the performance when the user disable some features. Specifically using CursorMoved is heavy so we can use this only when either `diagnostics_echo_cursor` or `highlight_references_enabled` is enabled.


mattn:
I know CursorMoved and CursorMovedI makes vim-lsp slow down. I don't think that this option should work with changing dyamically. Thanks.

mattn:
If someone create issues to use this options dynamically, let's discuss whether enable it.

PR #701, opened 2020-02-05
Remove prefix

mattn:
In selecting candidates in popup, if the item does not plain-text insertText  (i.e. label or insertText with snippet), it incude prefix text which already inserted. And it shoud be removed.

before
terminal7

after
terminal7

I confirmed html-language-server and metals. @hrsh7th Could you please check another Language Servers? Please note that this must use latest vim-lsp-settings, asyncomplete, asyncomplete-lsp since this feature need to set correct `typed_pattern` and `refresh_pattern`.

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -238,15 +238,24 @@ function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     let l:server_name = get(a:, 1, '')
 
     let l:word = ''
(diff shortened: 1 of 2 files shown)


mattn:
Sorry, I realized that we should not add code using typed_pattern. We should implement code to determine startcol like asyncomplete.vim. I'll send pull-request to:

* remove code of typed_pattern.
* add code to determine startcol similar to asyncomplete.

PR #703, opened 2020-02-07
Extension: Add support for clangd switch header/source

kent-tri:


mattn:
Thanks your contribution. But textDocument/switchHeader is not appear in standard language server specification. So I think we should not add this feature.

thomasfaingnaert:
As @mattn said, this is specific to one language server, and we would like vim-lsp to stay as LSP server agnostic as possible.

@mattn Can we move things like this to https://github.com/mattn/vim-lsp-settings?

kent-tri:
@mattn @thomasfaingnaert makes sense to me and I agree. I'll close this out and when I get a free minute will think more about how best to build a plugin system either in `vim-lsp-settings` or as a separate project that would be a better place to store these service-specific extensions.

mattn:
I will consider to add this positively.

PR #705, opened 2020-02-07
Don't allow hover window to close preview window

clktmr:
If g:lsp_preview_float is enabled there is no reason to close the
preview window on LspHover and LspSignatureHelp. Moreover the preview
window may be in use by other plugins or by the user.


mattn:
LGTM

mattn:
Thank you

PR #718, opened 2020-02-14
Change text edit implementation to support \r

hrsh7th:
This PR aims to support `\r\n`, `\r` and `\n` in text edits.

The approach is using only `\n` and modification uses `setline`.

### DIFF
--- a/autoload/lsp/utils.vim
+++ b/autoload/lsp/utils.vim
@@ -320,10 +320,10 @@ function! lsp#utils#make_valid_word(str) abort
    return l:str
 endfunction
 
-function! lsp#utils#split_by_eol(text) abort
+function! lsp#utils#_split_by_eol(text) abort
     let l:text = a:text
-    let l:text = substitute(l:text, "\r\n", "\n", '')
-    let l:text = substitute(l:text, "\r", "\n", '')
+    let l:text = substitute(l:text, "\r\n", "\n", 'g')
+    let l:text = substitute(l:text, "\r", "\n", 'g')
     return split(l:text, "\n", v:true)
 endfunction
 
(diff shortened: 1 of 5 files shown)


hrsh7th:
This implementation was ported by `hrsh7th/vim-vital-vs`.
It passed vscode's text edit test cases.

This PR should fix https://github.com/prabirshrestha/vim-lsp/issues/659.

hrsh7th:
I think it is ok to merge.

autoload/lsp/utils.vim
```diff
@@ -319,3 +319,11 @@ function! lsp#utils#make_valid_word(str) abort
    endif
    return l:str
 endfunction
+
+function! lsp#utils#_split_by_eol(text) abort
+    let l:text = a:text
+    let l:text = substitute(l:text, "\r\n", "\n", 'g')
+    let l:text = substitute(l:text, "\r", "\n", 'g')
+    return split(l:text, "\n", v:true)
```
mattn: Equivarent to this?

```
return split(a.text, '\r\n\|\r', 1)
```

hrsh7th: Thanks for your review.
test case can pass with `split(a:text, '\r\n\|\r\|\n', v:true)`.

I will change to it.

hrsh7th:
This approach is explained following.

```
abcdefg
hijklmn
opqrstu
```

If text edit is below

- range: `efg -> opq`
- newText: "000\n111\n222"

new implementation will create `l:new_lines` like below

```
abc000
111
222rstu
```

and then append `l:new_lines` simply.

mattn:
Thank you.

Issue #719, opened 2020-02-14
LspHover issue with Conceal


habamax:
Currently LspHover use conceal + cchar feature to hide `\` and for escaped chars:

![image](https://user-images.githubusercontent.com/234774/74512434-4ba36e80-4f19-11ea-9f3b-92a5137a094f.png)

As a followup of using cchar it also clears `Conceal` highlighting which breaks it for other filetypes.

![image](https://user-images.githubusercontent.com/234774/74512690-e3a15800-4f19-11ea-8afb-92b19a455767.png)

The code in question is:

https://github.com/prabirshrestha/vim-lsp/blob/master/syntax/lsp-hover.vim#L49

habamax:
What I suggest is to use another approach:

```
    if has('conceal')
        for l:escaped_char in ['`', '*', '_', '{', '}', '(', ')', '<', '>', '#', '+', '.', '!', '-']
            execute printf('syntax region vimLspMarkdownEscape matchgroup=Conceal start="\\\ze[%s]" end="[%s]\zs" concealends', l:escaped_char, l:escaped_char)
        endfor
    end

    hi link vimLspMarkdownEscape Pmenu
```

Which makes:

1. all those escaped chars to be highlighted as Pmenu -- thus it wouldn't have different background
2. will not break Conceal highlight group defined by colorscemes

![image](https://user-images.githubusercontent.com/234774/74512943-6cb88f00-4f1a-11ea-93c2-7d2db7820597.png)

habamax:
@prabirshrestha, @mattn should I create PR?

clason:
You can disable this altogether by registering the server with `'config': {  'hover_conceal': 0 }` .

habamax:
> 
> 
> You can disable this altogether by registering the server with `'config': { 'hover_conceal': 0 }` .

I do not see it in the code though...

`s:cleanup_markdown` is called in any way

clason:
Nevertheless, that's what I'm using ;)

See https://github.com/prabirshrestha/vim-lsp/blob/52539a54ae338993460205b4528d34caf175582e/autoload/lsp/ui/vim/output.vim#L327
and https://github.com/prabirshrestha/vim-lsp/blob/fabb572f7e85baaeeb57d8321ad3a901a1c5cf5a/doc/vim-lsp.txt#L801

habamax:
> Nevertheless, that's what I'm using ;)

Again, `Conceal` highlight group is explicitly cleared.

[...]

PR #720, opened 2020-02-14
Resolve LspHover Conceal Issue

habamax:
1. Do not clear Conceal highlight group
2. Make  `\` concealed using regions. This fixes different background
   of concealed cchar and popup window original solution had.

See #719 for details.

### DIFF
--- a/syntax/lsp-hover.vim
+++ b/syntax/lsp-hover.vim
@@ -41,12 +41,9 @@ function! s:cleanup_markdown() abort
     " Workaround for: https://github.com/palantir/python-language-server/issues/386
     if has('conceal')
         for l:escaped_char in ['`', '*', '_', '{', '}', '(', ')', '<', '>', '#', '+', '.', '!', '-']
-            execute printf('syntax match markdownEscape "\\[][%s]" conceal cchar=%s', l:escaped_char, l:escaped_char)
+            execute printf('syntax region vimLspMarkdownEscape matchgroup=Conceal start="\\\ze[%s]" end="[%s]\zs" concealends', l:escaped_char, l:escaped_char)
         endfor
     end
-
-    " Don't highlight concealed chars
-    highlight clear Conceal
 endfunction
 
 call s:do_highlight()


syntax/lsp-hover.vim
```diff
@@ -41,12 +41,11 @@ function! s:cleanup_markdown() abort
...
     if has('conceal')
         for l:escaped_char in ['`', '*', '_', '{', '}', '(', ')', '<', '>', '#', '+', '.', '!', '-']
-            execute printf('syntax match markdownEscape "\\[][%s]" conceal cchar=%s', l:escaped_char, l:escaped_char)
+            execute printf('syntax region vimLspMarkdownEscape matchgroup=Conceal start="\\\ze[%s]" end="[%s]\zs" concealends', l:escaped_char, l:escaped_char)
         endfor
     end
 
-    " Don't highlight concealed chars
-    highlight clear Conceal
+    highlight link vimLspMarkdownEscape Pmenu
```
thomasfaingnaert: I'm using a custom background for `LspHover`s, so with this change, it looks like this:
image

Instead of clearing `Conceal`, maybe you can add a new highlight group `LspConceal` that is cleared instead?

habamax: Is it only the background issue? or a italic?

thomasfaingnaert: Only the background. Italic is because pyls is not escaping the `_`.

habamax: How did you set up LspHover background?

thomasfaingnaert: I have added it to the docs here: https://github.com/prabirshrestha/vim-lsp/blob/master/doc/vim-lsp.txt#L264-L279

habamax: 1. So, we shouldn't globally clear `Conceal` in any case.
2. Current implementation uses `cchar` which uses `Conceal` highlight group
3. Is there a way to query vim/nvim popup window for highlighting it uses? To be able to use it instead of `Pmenu`?

habamax: Huh! It is easy -- we don't have to define highlighting for the `vimLspMarkdownEscape`!

thomasfaingnaert: Getting the option `'wincolor'` or `'winhighlight'` should work. But maybe it's easier to define another highlight group without properties, so that it's "transparent"?

thomasfaingnaert: Exactly :smile:

habamax: Could you try again (I have force pushed changes)

thomasfaingnaert:
Seems to work well, I'll merge it in. Thanks for the contribution!

habamax:
Thank you!

PR #724, opened 2020-02-17
Fix cursor pos after text edit

hrsh7th:
Sometimes, target_bufname is not equaled to current_bufname.


hrsh7th:
I'm sorry.
CI failed but I merge it.
It seems CI failed by neovim's release.

PR #729, opened 2020-02-25
Export root_uri resolved

mattn:
Current implementation does not provide way know there is the root_uri which is resolved.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -85,6 +85,10 @@ function! lsp#get_server_info(server_name) abort
     return s:servers[a:server_name]['server_info']
 endfunction
 
+function! lsp#get_server_root_uri(server_name) abort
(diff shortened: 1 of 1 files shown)


autoload/lsp.vim
```diff
@@ -516,16 +520,14 @@ function! s:ensure_init(buf, server_name, cb) abort
     let l:server_info = l:server['server_info']
     if has_key(l:server_info, 'root_uri')
         let l:root_uri = l:server_info['root_uri'](l:server_info)
-    else
-        let l:root_uri = lsp#utils#get_default_root_uri()
     endif
 
     if empty(l:root_uri)
```
hrsh7th: This could cause error.
`l:root_uri` should initialize in above of this line, I think.

mattn: Right

PR #734, opened 2020-02-29
Call command after edit

mattn:
codeAction may contains edit and command both

https://microsoft.github.io/language-server-protocol/specifications/specification-current/

Language Server Specification says:

```
	/**
	 * The workspace edit this code action performs.
	 */
	edit?: WorkspaceEdit;

	/**
	 * A command this code action executes. If a code action
	 * provides an edit and a command, first the edit is
	 * executed and then the command.
	 */
	command?: Command;
```


hrsh7th:
It' my bad. thanks! I will merge it when CI passed.

PR #739, opened 2020-03-05
Compute overflow information in completion

hrsh7th:
This PR is an experimental feature, please verify it has no edge case.

I had been investigating VSCode's completion behavior.
In this investigation, I found strange behavior.

### json-languageserver's completion behavior

I test this situation of completion in `package.json`. (`"` is automatically closed by lexima.vim)

```json
{
  "na|"
}
```

I confirm `name` candidates and then `VSCode` gets below results.

```json
{
  "name": "|"
}
```

This is an ideal behavior, I think. (`"` was not duplicates)
But vim-lsp or other client does not get this result.

My investigation founds the point of this behavior in VSCode's implementation.

https://github.com/microsoft/vscode/blob/master/src/vs/editor/contrib/suggest/suggestController.ts#L339

This PR implements VSCode's `getOverwriteInfo` maybe.

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -250,6 +250,7 @@ endfunction
 
 function! lsp#omni#default_get_vim_completion_item(item, ...) abort
     let l:server_name = get(a:, 1, '')
+    let l:complete_position = get(a:, 2, lsp#get_position())
 
     let l:word = ''
     if get(a:item, 'insertTextFormat', -1) == 2 && !empty(get(a:item, 'insertText', ''))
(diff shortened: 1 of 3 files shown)


hrsh7th:
This is not working when set `iskeyword+=\"`.
I will more deeply investigate it.

hrsh7th:
Finally, I did implement the completion that LSP spec-compliant maybe.

I tested it with below cases.

- html-languageserver
    - attribute completion (`class|>` -> `class="|">`)
    - indent fixer completion (`    </|>` -> `  </div|>`)

- intelephense
    - method completion (`$this->set|` -> `$this->set_value(|)`

- json-languageserver
    - property completion (`"n|"` -> `"name": "|"`)

hrsh7th:
mattn tested it with metals.

The metals returns bit rare textEdit but I heard it works.
Thanks @mattn!

I think this is very useful so I merge this now.
If someone sent an issue, I will try to solve the issue.

PR #764, opened 2020-03-24
Follow vscode textEdit behavior

hrsh7th:
This PR aims to fix text edit behavior.

I found the problem when using rust-analyzer.
rust-analyzer seems to expect to fix cursor col after applying text edit and VSCode seems to do it.

Before
follow-vscode-behavior-before

After
follow-vscode-behavior-after

I thought vim-lsp should supports `virtual edit` position.
But VSCode and LSP does not tell how to support it.

My previous implementation trying to support it but I removed those codes from this PR.

--- Updated ---

New text edit implementation fix cursor position strictly so I removed winsaveview and winrestview.

### DIFF
--- a/autoload/lsp/utils/text_edit.vim
+++ b/autoload/lsp/utils/text_edit.vim
@@ -1,21 +1,18 @@
 function! lsp#utils#text_edit#apply_text_edits(uri, text_edits) abort
     let l:current_bufname = bufname('%')
     let l:target_bufname = lsp#utils#uri_to_path(a:uri)
-    let l:cursor_pos = getpos('.')[1 : 3]
+    let l:cursor_position = lsp#get_position()
     let l:cursor_offset = 0
     let l:topline = line('w0')
 
     call s:_switch(l:target_bufname)
     for l:text_edit in s:_normalize(a:text_edits)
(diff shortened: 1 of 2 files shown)


autoload/lsp/utils/text_edit.vim
```diff
@@ -1,29 +1,26 @@
...
-        let l:cursor_offset += s:_apply(bufnr(l:target_bufname), l:text_edit, l:cursor_pos)
+        let l:cursor_offset += s:_apply(bufnr(l:target_bufname), l:text_edit, l:cursor_position)
     endfor
     call s:_switch(l:current_bufname)
 
     if bufnr(l:current_bufname) == bufnr(l:target_bufname)
-        let l:length = strlen(getline(l:cursor_pos[0])) + 1
-        let l:cursor_pos[2] = max([0, l:cursor_pos[1] + l:cursor_pos[2] - l:length])
-        let l:cursor_pos[1] = min([l:length, l:cursor_pos[1] + l:cursor_pos[2]])
-        call cursor(l:cursor_pos)
```
hrsh7th: Those code trying to support `virtual col` but I've removed it.

PR #782, opened 2020-04-07
Replace builtin trim() by lsp#utils#_trim()

jagd:
Builtin trim() does not exist in e.g. vim 8.0, which is the default version of Ubuntu 18.04 LTS.


hrsh7th:
Thank you for the fix.

PR #797, opened 2020-04-21
Add example for g:lsp_get_vim_completion_item

clktmr:
I added an example because there are some pitfalls if one is not fluent
in vimscript:
- using call() instead of :call
- passing a:item and varargs correctly
- variable name for FuncRef must be uppercase
- g:lsp_get_vim_completion_item must be a List

### DIFF
--- a/doc/vim-lsp.txt
+++ b/doc/vim-lsp.txt
@@ -529,6 +529,16 @@ g:lsp_get_vim_completion_item               *g:lsp_get_vim_completion_item*
     Note: You can reuse functionality provided by vim-lsp by calling
     `lsp#omni#default_get_vim_completion_item` from within your function.
 
+    Example: >
+    function! LspGetCompletionItem(item, ...) abort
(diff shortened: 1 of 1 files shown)


prabirshrestha:
Merged. Thanks.

prabirshrestha:
@clktmr  I'm looking into deprecating `g:lsp_get_vim_completion_item` but before I go ahead with it I would like to understand if you still need this and for what are you using it.

This will be removed once this https://github.com/prabirshrestha/vim-lsp/pull/835 gets merged.

clktmr:
I mainly used this as a hack against verbosity of some servers, e.g.:
- Filtering completion items by kind, e.g. in Lua every item has a
  duplicate of kind `snippet`.  Currently this can't be done with
  `vim-lsp-completion-filter`.
- Reducing or limiting completion items in length, e.g. `clangd` will
  show template and function parameters and one long item will bloat the
  whole popup-menu.

But a performance optimization is much appreciated and I will gladly
give up on this feature for it.

PR #830, opened 2020-05-28
Fix var names in help about g:lsp_diagnostics_float_delay

a5ob7r:
Not lsp_diagnostics_echo_delay but lsp_diagnostics_float_delay.


hrsh7th:
Thank you for fix it.

PR #866, opened 2020-07-12
fix document in PERFORMANCE section

mopp:
## What is the problem
The variable name was different from the implementation 
https://github.com/prabirshrestha/vim-lsp/blob/11b13507002e42adf77b35ed63b4164ecc5d5b8a/plugin/lsp.vim#L6

## What I did
I just copied and pasted the actual code into the document.


mattn:
Good catch. Thank you.

PR #873, opened 2020-07-29
Trigger didOpen for buffers already loaded

mattn:
When using `-p` flag, BufReadPost events are not triggered since those are already triggered.

```
$ vim -p *
```

```
:bnext
```


hrsh7th:
LGTM

PR #878, opened 2020-08-06
Handle invalid range

mattn:
Some Language Server return invalid range for textDocument/definition. This change ignore the invalid range.


hrsh7th:
I think it has not side-effect so it ready to merge.

PR #925, opened 2020-11-04
Add missing arg for <Plug>(lsp-workspace-symbol)

### LINKED ISSUE #924: <Plug>(lsp-workspace-symbol) throws E119
Using the `<Plug>(lsp-workspace-symbol)` mapping throws `E119: Not enough arguments for function: lsp#ui#vim#workspace_symbol`.

The function requires one argument, but the mapping doesn't provide any.
Presumably it should be called with an empty string argument so that the function prompts for a query.

seandewar:
Closes #924.


mattn:
Thank you.

PR #938, opened 2020-11-20
Add support for codeLens/resolve

dradtke:
This PR updates the code lens action to invoke `codeLens/resolve` on each result. According to the [spec](https://microsoft.github.io/language-server-protocol/specification#codeLens_resolve), the result of `textDocument/codeLens` is partial. In particular, the `command` value is missing, and an explicit call to resolve the lens is required in order to populate it.

### DIFF
--- a/autoload/lsp/ui/vim/code_lens.vim
+++ b/autoload/lsp/ui/vim/code_lens.vim
@@ -137,7 +137,6 @@ function! s:handle_code_lens_resolve(ctx, server_name, command_id, sync, bufnr, 
 endfunction
 
 function! s:handle_one_code_lens(server_name, sync, bufnr, code_lens) abort
-  let g:executed_code_lens = a:code_lens
     call lsp#ui#vim#execute_command#_execute({
     \   'server_name': a:server_name,
     \   'command_name': get(a:code_lens['command'], 'command', ''),
--- a/autoload/lsp/ui/vim/code_lens.vim
+++ b/autoload/lsp/ui/vim/code_lens.vim
@@ -77,6 +77,53 @@ function! s:handle_code_lens(ctx, server_name, command_id, sync, bufnr, data) ab
     endif
     call lsp#log('s:handle_code_lens', l:total_code_lenses)
 
+    let l:resolve_ctx = {
+    \ 'count': len(l:total_code_lenses),
+    \ 'results': [],
+    \}
+
+    for l:code_lens in l:total_code_lenses
+      call lsp#send_request(l:server_name, {
+            \ 'method': 'codeLens/resolve',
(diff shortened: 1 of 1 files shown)


prabirshrestha:
Can you provide sample instructions so that I could try it out.

prabirshrestha:
@dradtke Do you have a sample lang server that I can try this on?

dradtke:
This was tested with eclipse.jdt.ls. Unfortunately I don't have a project handy for a demonstration, as I mostly just use it at work.

-- 
  Damien Radtke
  me@damienradtke.com

prabirshrestha:
@dradtke I can try it with my own project. Can you add steps for me to try out and expected behavior?

prabirshrestha:
I tried both rust-analyzer and eclipse.jdt.ls and there are few things I noticed.

1. Server send `codelensProvider.resolveProvider = true` if it supports resolve calls. This means we need to add `lsp#capabilities#has_code_lens_resolve_provider(server_name)`  and only send resolve call if it supports.
2. Based on the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeLens). We should only call resolve when command doesn't exist.

```typescript
/**
 * A code lens represents a command that should be shown along with
 * source text, like the number of references, a way to run tests, etc.
 *
 * A code lens is _unresolved_ when no command is associated to it. For
 * performance reasons the creation of a code lens and resolving should be done
 * in two stages.
 */
interface CodeLens {
	/**
	 * The range in which this code lens is valid. Should only span a single
	 * line.
	 */
	range: Range;

	/**
	 * The command this code lens represents.
	 */
	command?: Command;

	/**
	 * A data entry field that is preserved on a code lens item between
	 * a code lens and a code lens resolve request.
	 */
	data?: any
}
```

In master eclipse langserver is broken while rust-analyzer is working. With this PR it seems that eclipse works but rust-analyzer is broken. Do note that both rust-analyzer and eclipse does seems to set `codelensProvider.resolveProvider` as `true`. The only difference being `rust-analyzer` seems to include `command` while eclipse doesn't.

Could you update the PR to support both 1 and 2?

dradtke:
Interesting, that's a good comparison. I can try to take a look at improving it soon, though no guarantees on when exactly that will be.

prabirshrestha:
Please give https://github.com/prabirshrestha/vim-lsp/pull/938 a try. It support both rust and eclipse.jdt.ls.

I also did notice that eclispse has few custom commands. So would be good to send those PRs to vim-lsp-settings similar to this https://github.com/mattn/vim-lsp-settings/pull/362

prabirshrestha:
#938 is merged and should handle the fix now.

PR #947, opened 2020-12-11
Fix highlight references not getting cleared properly

subnut:
`matchadd()` and `matchdelete()` are window-local functions, and the matches are
window-local. So, use window-local variable to store and delete them.

Otherwise, the highlights are often not deleted properly, and remain
fixed. <strike>(The issue is not consistently reproducible, but frequent)</strike>

EDIT: I was able to reproduce it consistently in a particular case. See GIF below in 3rd comment

### DIFF
--- a/autoload/lsp/internal/highlight_references.vim
+++ b/autoload/lsp/internal/highlight_references.vim
@@ -96,8 +96,13 @@ function! s:set_highlights(data) abort
     if s:in_reference(l:position_list) == -1 | return | endif
 
     " Store references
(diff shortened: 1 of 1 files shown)


subnut:
The failing test seems to be unrelated. Please check.

subnut:
# Before (notice the residual highlight)

gifcast_201211085805

# After (notice how it goes away)

gifcast_201211090151

prabirshrestha:
Merged. Thanks.

PR #955, opened 2020-12-22
do not use vim popup if not available

prabirshrestha:
fix https://github.com/prabirshrestha/vim-lsp/issues/952


prabirshrestha:
closing. will send proper fix in new pr.

PR #989, opened 2020-12-30
Fix tests

mattn:


mattn:
@prabirshrestha Fixed some failing tests. But I can't figure out macos.

mattn:
Failing on older version of nvim.
image

prabirshrestha:
This looks good. I'm merging it. I think we should remove integration tests. Wait apparently doesn't work all the time. I wrote few other integration tests but  it got stuck so never worked. We need vim-themis to support async tests. https://github.com/thinca/vim-themis/issues/61

PR #1001, opened 2021-01-02
Improve error message at unsupported feature

rhysd:
This change will work fine even if the filetype is empty:

e.g.

```
Renaming not supported for filetype ''
```

Current implementation shows an error message like below:

```
Renaming not supported for 
```


mattn:
Thank you

PR #1028, opened 2021-01-08
Fix #1024

### LINKED ISSUE #1024: Markdown documentation is not wrapped
Hi, I wonder if it is possible to wrap documentation for the methods to fit the popup window if it is in the markdown style?
Right now I have max width set to `79`, but it seems to have no effect, or am I missing something?

```vim
let g:lsp_preview_max_width = 79
```

Here's what it looks like:

image

subnut:
Closes prabirshrestha/vim-lsp#1024


hrsh7th:
LGTM.

hrsh7th:
I think we should align to `output.vim`'s behavior but I think it's ok to merge for now.

PR #1042, opened 2021-01-12
Fix/showsigpreview

jerojasro:
Previously, when using the preview window to show signature information, the code failed with an stacktrace like:

     "s:on_stdout client request on_notification() error",
     "Vim(return):E121: Undefined variable: l:bufferlines",
     "function <SNR>118_out_cb[2]..
     <SNR>107_on_stdout[79]..
     <SNR>92_request_on_notification[3]..
     <SNR>93_createNext[1]..
     <SNR>93_subscribeSourceCallback[2]..
     <lambda>738[1]..
     <SNR>94_handle_signature_help[48]..
     lsp#ui#vim#output#preview[39]..
     lsp#ui#vim#output#get_size_info, line 22"

This PR ensures that the required variable is properly defined in all the situations (popup, nvim, preview window)

### DIFF
--- a/autoload/lsp/ui/vim/output.vim
+++ b/autoload/lsp/ui/vim/output.vim
@@ -303,7 +303,6 @@ function! lsp#ui#vim#output#get_size_info(winid) abort
     endif
 
     return [l:bufferlines, l:maxwidth]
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #1055, opened 2021-01-20
Documentation popup: fix size/markdown/detail

hrsh7th:
This PR is continuing work of `completion-resolve-documentation` branch.

1. Add `markdown` for client capabilities.
2. Fix documentation window positions
3. Support `detail` field. (It will be useful with gopls)

### DEMO

https://user-images.githubusercontent.com/629908/105219109-66ef1300-5b99-11eb-872e-5f4ff509679f.mp4


autoload/lsp.vim
```diff
@@ -475,7 +475,7 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \       'completion': {
     \           'dynamicRegistration': v:false,
     \           'completionItem': {
-    \              'documentationFormat': ['plaintext'],
+    \             'documentationFormat': ['markdown'],
```
prabirshrestha, line 478: since array is ordered can we instead use `['markdown', 'plaintext']`

autoload/lsp/internal/completion/documentation.vim
```diff
@@ -65,43 +68,113 @@ endfunction
...
     endif
+
+    return [a:event.row, l:col]
+endfunction
+
+function! s:get_doc_win() abort
+    if exists('s:doc_win')
+        return s:doc_win
+    endif
+    if !s:FloatingWindow.is_available()
```
prabirshrestha, line 168: check not required since we check it during `_enable`

autoload/lsp/internal/completion/documentation.vim
```diff
@@ -65,43 +68,113 @@ endfunction
...
+        endif
+    endfunction
+    call timer_start(1, { -> l:ctx.callback(a:force) })
 endfunction
 
-function! s:close_floating_window() abort
-    if exists('s:floating_win')
-        call s:floating_win.close()
+function! s:compute_position(event, size) abort
+    let l:col_if_right = a:event.col + a:event.width + 1 + (a:event.scrollbar ? 1 : 0)
```
prabirshrestha, line 145: not required now and can do later. but we should port the logic from neovim doc floating. it solves lot of overlapping issue and also includes docking. we can slowly do this as separate PR.

prabirshrestha: One way to try out is by trying to open popups around corner or in small splits and windows or bottom and top and see if it overlaps or not. there is also dock support by using popup which some folks that have tall monitors use but we can fix this later.

autoload/lsp/internal/completion/documentation.vim
```diff
@@ -65,43 +68,113 @@ endfunction
 
 function! s:show_floating_window(event, managed_user_data) abort
     let l:completion_item = a:managed_user_data['completion_item']
-    let l:documentation = get(l:completion_item, 'documentation', '')
-    if type(l:documentation) == type({}) && has_key(l:documentation, 'value')
-        let l:documentation = l:documentation['value']
+
+    " Ignore if documentation does not exists.
+    if !has_key(l:completion_item, 'documentation')
```
prabirshrestha, line 73: should this also check for detail? I'm wondering if we should show detail if documentation doesn't exist. i don't have a strong preference for this. Might partially solve this issue https://github.com/prabirshrestha/vim-lsp/issues/1044

prabirshrestha: currently seems like we need to have documentation in order to show details + doc. not sure how vs-code does it.

autoload/lsp/internal/completion/documentation.vim
```diff
@@ -65,43 +68,113 @@ endfunction
...
-    if exists('s:floating_win')
-        call s:floating_win.close()
+    " Add detail field if provided.
+    if type(get(l:completion_item, 'detail', v:null)) == type('')
+        if !empty(l:completion_item.detail)
+            let l:detail = s:MarkupContent.normalize({
+            \   'language': &filetype,
+            \   'value': l:completion_item['detail'],
+            \ })
+            let l:documentation = l:detail . "\n\n" . l:documentation
```
prabirshrestha, line 92: do we need to check for empty detail?

prabirshrestha (approved):
Looks great. most are small nit comments so feel free to fix in this PR or in other PRs and merge in my branch,

prabirshrestha:
Noticed a bug where popup is not getting rendered correctly. Only impacts vim.
image

Also noticed that if I open the doc popup and try to close vim it tells me I haven't saved a buffer which points to the last vim popup.

autoload/lsp/internal/completion/documentation.vim
hrsh7th: _I personally prefer to avoid displaying popup-menu if `left/right` has no enough space for both._

But agree with you. we can improve it later.

hrsh7th:
Hm... the bug is the dark bar that placed after `relations`, right?

Yes. I noticed the some of bugs. I will fix it later.

hrsh7th:
The scroll feature.

```
nnoremap <expr> <C-f> <SID>scroll(+4)
nnoremap <expr> <C-u> <SID>scroll(-4)
xnoremap <expr> <C-f> <SID>scroll(+4)
xnoremap <expr> <C-u> <SID>scroll(-4)
inoremap <expr> <C-f> <SID>scroll(+4)
inoremap <expr> <C-u> <SID>scroll(-4)

function! s:scroll(count) abort
  let l:ctx = {}
  function! l:ctx.callback() abort closure
    let l:Window = vital#vital#import('VS.Vim.Window')
    for l:winid in l:Window.find({ winid -> l:Window.is_floating(winid) })
      let l:info = l:Window.info(l:winid)
      call l:Window.scroll(l:winid, l:info.topline + a:count)
    endfor
  endfunction
  call timer_start(0, { -> l:ctx.callback() })
  return "\<Ignore>"
endfunction
```

autoload/lsp.vim
hrsh7th: Thank you. Fixed.

autoload/lsp/internal/completion/documentation.vim
hrsh7th: I've removed it.

autoload/lsp/internal/completion/documentation.vim
hrsh7th: I've improved to resolve `documentation/detail`.

autoload/lsp/internal/completion/documentation.vim
hrsh7th: It is not need to check. because `MarkupContent.normalize({ `language': ..., 'value': ... })` will creates fenced code block (so it always has contents.)

hrsh7th:
This PR does not supports the below features.

1. scrolling
- IMO, support in another PR is better.

2. docked
- It's a bit difficult.... Hm...

3. lsp_float_opened
- I will support it in this PR.

prabirshrestha:
Thanks I have merged your branch.

 For docked -> Can we not use the preview window? I personally don't use this so not sure we really need it. What we currently have is so much better so not worried about it for now. We can add it back in future if there is demand. This simplifies our code too.

scroll can come in different PR.

prabirshrestha:
@hrsh7th the bug is about the dark bar after relations. wondering if we never had this issue in the popup before because vim had borders so it always took full width. wondering if it has to do something with transparency.

prabirshrestha:
haven't played around but seems like there is `:h popup-mask`

PR #1095, opened 2021-03-13
Use 'label' value when 'insertText' or 'filterText' is falsy

rhysd:
From [LSP v3.1.6](https://microsoft.github.io/language-server-protocol/specification),

```
	/**
	 * A string that should be used when filtering a set of
	 * completion items. When `falsy` the label is used as the
	 * filter text for this item.
	 */
	filterText?: string;

	/**
	 * A string that should be inserted into a document when selecting
	 * this completion. When `falsy` the label is used as the insert text
	 * for this item.
	 *
	 * The `insertText` is subject to interpretation by the client side.
	 * Some tools might not take the string literally. For example
	 * VS Code when code complete is requested in this example
	 * `con<cursor position>` and a completion item with an `insertText` of
	 * `console` is provided it will only insert `sole`. Therefore it is
	 * recommended to use `textEdit` instead since it avoids additional client
	 * side interpretation.
	 */
	insertText?: string;
```

the label should be used as filter text and insert text when they are falsy value. I noticed that vim-lsp does not consider the case where `insertText` is falsy in `s:on_complete_done_after()` at `completion.vim`. And I found that vim-lsp does not use label when `filterText` is falsy.

This PR fixes the issue. And reverts #1093 and #1094.

zls returns `null` as the falsy value of the `insertText` entry. `s:on_complete_done_after()` did not handle the `null` value correctly and it generated invalid `TextEdit` value with null `newText` entry. This was actually cause of #1093.

I tried to add tests for this. But I could not find test for `filterText` and `s:on_complete_done_after` so I could not add them. Please let me know if I'm missing something about tests.

### DIFF
--- a/autoload/lsp/ui/vim/completion.vim
+++ b/autoload/lsp/ui/vim/completion.vim
@@ -107,7 +107,11 @@ function! s:on_complete_done_after() abort
     else
       let l:overflow_before = 0
       let l:overflow_after = 0
-      let l:text = get(l:completion_item, 'insertText', l:completed_item['word'])
+      let l:text = get(l:completion_item, 'insertText', '')
+      if empty(l:text)
+        " When insertText is `falsy` the label is used as the insert text
+        let l:text = get(l:completion_item, 'label', l:completed_item['word'])
+      endif
     endif
 
     " apply snipept or text_edit
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -93,8 +93,15 @@ endfunction
 
 function! s:get_filter_label(item) abort
     let l:user_data = lsp#omni#get_managed_user_data_from_completed_item(a:item)
-    if has_key(l:user_data, 'completion_item') && has_key(l:user_data['completion_item'], 'filterText') && !empty(l:user_data['completion_item']['filterText'])
-        return lsp#utils#_trim(l:user_data['completion_item']['filterText'])
+    if has_key(l:user_data, 'completion_item') && has_key(l:user_data['completion_item'], 'filterText')
+        let l:filter_text = l:user_data['completion_item']['filterText']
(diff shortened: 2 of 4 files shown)


autoload/lsp/ui/vim/completion.vim
```diff
@@ -107,7 +107,11 @@ function! s:on_complete_done_after() abort
     else
       let l:overflow_before = 0
       let l:overflow_after = 0
-      let l:text = get(l:completion_item, 'insertText', l:completed_item['word'])
+      let l:text = get(l:completion_item, 'insertText', '')
+      if empty(l:text)
+        " When insertText is `falsy` the label is used as the insert text
+        let l:text = get(l:completion_item, 'label', l:completed_item['word'])
+      endif
     endif
```
hrsh7th, line 115: Thank you.
In spec, the insertText can't be `null` but some servers seem to return it as null

rhysd: Hmm, I'm not familiar with the specification. But it says insertText can be 'falsy'. I thought `null` is included in the falsy values. Anyway `empty()` can handle both empty string and null so it should work well even if null is invalid for insertText.

hrsh7th: Ah. You are right.

Sorry. I have overlooked doc comments. (I had only seen type annotations).

hrsh7th (approved):

hrsh7th:
I checked it with vscode-html-language-server/rust-analyzer/vim-language-server/intelephense/

rhysd:
FYI: LSP maintainer kindly clarified that `null` cannot be assigned to `insertText` since the field is typed as `?insertText: string`. Only `""` and `undefined` can be falsy values for the field.

https://github.com/zigtools/zls/pull/251

PR #1097, opened 2021-03-14
Fix calculating last line of text diff in Lua and add new CI for Vim on macOS

rhysd:
Hi,

This PR

- adds CI workflow for Vim on macOS. I used [MacVim](https://github.com/macvim-dev/macvim) which is most popular Vim distribution for macOS
- fixes calculating text diff in Lua

I confirmed all tests pass with the new workflow on my forked repository: https://github.com/rhysd/vim-lsp/runs/2107026701?check_suite_focus=true

### DIFF
--- a/.github/workflows/mac_vim.yml
+++ b/.github/workflows/mac_vim.yml
@@ -1,0 +1,36 @@
+name: mac_vim
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


rhysd:
Hmm, the change makes Neovim CI workflows fail. I'll take a look.

rhysd:
Note: I'm not understanding why offset is different between Vim and Neovim

https://github.com/prabirshrestha/vim-lsp/blob/6c21eb65da57cbe0c580b2f26e331872901795d3/autoload/lsp/utils/diff.vim#L100

rhysd:
@prabirshrestha @hrsh7th @mattn Now this PR is ready. Would someone review this?

PR #1101, opened 2021-03-15
Fix CodeAction sign

hrsh7th:
This PR aims to avoid always showing CodeAction signs if the server returns `source.organizeImports` as CodeAction response.


mattn:
Thank you.

PR #1108, opened 2021-03-16
Add integration tests with gopls instead of rust-analyzer

rhysd:
#1098 removed integration tests from CI because they are flaky. However, checking integration between client and real-world servers is important. This PR tries to restore running integration tests on CI.

I investigated the reason of flakiness and understood that `rust-analyzer` takes longer time than timeout (10 seconds). For heavy analysis, `rust-analyzer` seem to need more than 10 seconds even for small projects.

Instead, I know that `gopls` has small startup time. So I reverted #1098 and replaced `rust-analyzer` with `gopls` for integration test cases. Changes of this PR are:

- Added `testproject-go` for integration tests using `:LspDocumentFormatSync`
- Implemented simpler helpers for gopls in `test/utils/autoload/lsp/test.vim`
- Replaced `test/integration/rust` with `test/integration/go`
- Installed `gopls` instead of `rust-analyzer` in all workflows. Previously integration tests were run on non-Windows workers before #1098 but this time I enabled integration tests in all workers since `gopls` also works fine on Windows

I confirmed all workflows pass.

https://github.com/rhysd/vim-lsp/actions?query=branch%3Aflaky-tests

One concern is that installing `gopls` takes a bit longer time than `rust-analyzer`. Since `gopls` does not provide pre-built binaries, we need to install it from source. It means we need to install Go toolchain for building `gopls` binaries and build the binary in each workflow. It slows workflows around 0.5~2 minutes.

Before (current HEAD of vim-lsp):

スクリーンショット 2021-03-17 3 25 56

After (this PR branch):

スクリーンショット 2021-03-17 3 26 12


rhysd:
@mattn I think you're expert of `gopls`. Please let me know if `gopls` pre-built binaries are released somewhere.

prabirshrestha (commented):
I'm ok going with golang for everything if it is faster.

I went with rust because I'm familiar with it and doesn't require any compilations and so on.

.github/workflows/linux_neovim.yml
```diff
@@ -31,21 +31,24 @@ jobs:
...
+        with:
+          go-version: '^1.16.1'
+      - name: Install gopls
         shell: bash
         run: |
-          mkdir -p ~/langservers
-          curl -L https://github.com/rust-analyzer/rust-analyzer/releases/download/2020-12-21/rust-analyzer-linux -o ~/langservers/rust-analyzer
-          chmod u+x ~/langservers/rust-analyzer
-          ~/langservers/rust-analyzer --version
+          go get golang.org/x/tools/gopls@latest
```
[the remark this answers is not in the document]
rhysd: I thought using the latest stable version is better. But if you prefer fixed versions, yes we can pin the versions. I'll try to specify them with environment variables.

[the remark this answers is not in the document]
rhysd: @prabirshrestha I pinned Go and gopls version through `VIM_LSP_GO_VERSION` and `VIM_LSP_GOPLS_VERSION` at https://github.com/prabirshrestha/vim-lsp/pull/1108/commits/eac78f904ccf7e4dd17c008139833e81766a5dee.

prabirshrestha:
About the perf we could try to have cache  but not required.

I would choose tests and stability over the speed as it isn't that bad.

rhysd:
> About the perf we could try to have cache but not required.

Yes, I thought using cache. However it's a bit difficult to determine when cache should be invalidated to me. So I want to consider it after this PR is merged as separate PR.

prabirshrestha:
for cache key i would use 3 things.

```
cacheversion = 1
cachekey = go version + gopls version + cacheversion
```

in case we make mistake we can always bump cache version or what i like to call the counter of how many mistakes we have made :)

test/utils/autoload/lsp/test.vim
```diff
@@ -31,6 +33,23 @@ function! lsp#test#openproject(name, options) abort
             \ lsp#callbag#take(1),
             \ lsp#callbag#toList(),
             \ ).wait({ 'timeout': 10000, 'sleep': 100 })
```
prabirshrestha, line 35: should we also do the same here as golang

open -> close -> wait -> bwipeout

im more than happy to remove rust then we don't need to worry about it.

test/testproject-go/documentformat.go
```diff
@@ -0,0 +1,4 @@
+package main
+func main() {
+print("hello, world!")
```
prabirshrestha, line 3: it will be good if we can have mutli byte chars.

prabirshrestha: alpha/beta are universal and easy for anyone coming from different language to speak.

https://github.com/prabirshrestha/vim-lsp/blob/master/test/testfiles/multibyte.txt

prabirshrestha (changes requested):

.github/workflows/linux_neovim.yml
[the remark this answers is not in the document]
prabirshrestha: I prefer pinning since the builds are reproducible.

A simple example that could break without pinning is might be the server today shows basic hover information but in future it may show more information. similar with diagnostics they may catch more warnings.

instead of going into a rabbit hole of why the test is failing, it becomes easier to know a test fails due to a specific gopls update.

i think it is worth it in the long run though we would have to bump the version now and then.

test/utils/autoload/lsp/test.vim
rhysd: I removed unused test helper for Rust project as much as possible at ca86fb4. Whole project directory and helpers could not be removed because `lsp#test#projectdir('rust')` is used by other test case.

test/testproject-go/documentformat.go
rhysd: Thank you for the missing point. I added multibyte texts in the test case at b81b79a.

rhysd:
> for cache key i would use 3 things.

Thank you for your advice. I'll try cache configuration. Many commits would be force-pushed to this branch for experiments.

rhysd:
I successfully tried adding cache at `neovim_linux` workflow.

スクリーンショット 2021-03-17 21 22 03

- No cache: 56s
- Cached: 26s

I'll implement same logic to all workflows

rhysd:
@prabirshrestha I completed adding cache to all workflows. Here is the result:

- No cache:

スクリーンショット 2021-03-17 21 43 19

- Cached:

スクリーンショット 2021-03-17 21 45 45

The longest job `mac_vim` is important. The time was much improved as follows:

- No cache: 4m54s
- Cached: 1m35s

mattn:
Sorry, I don't know whether the gopls binaries are released somewhere.

rhysd:
No problem, thank you for the answer. I think using cache is sufficient in terms of speed.

rhysd:
I rebased this branch onto the latest master.

prabirshrestha:
Thanks for the PR. This is going to be a huge improvement.

Once we start having more tests we can start enforcing it.

this will also make me more confident as I keep refactoring internals for perf.

PR #1115, opened 2021-03-21
Handle disabled code action

rhysd:
[LSP v3.16](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) defines `disabled` optional property to response of CodeAction request.

```
	/**
	 * Marks that the code action cannot currently be applied.
	 *
	 * Clients should follow the following guidelines regarding disabled code
	 * actions:
	 *
	 * - Disabled code actions are not shown in automatic lightbulbs code
	 *   action menus.
	 *
	 * - Disabled actions are shown as faded out in the code action menu when
	 *   the user request a more specific type of code action, such as
	 *   refactorings.
	 *
	 * - If the user has a keybinding that auto applies a code action and only
	 *   a disabled code actions are returned, the client should show the user
	 *   an error message with `reason` in the editor.
	 *
	 * @since 3.16.0
	 */
	disabled?: {

		/**
		 * Human readable description of why the code action is currently
		 * disabled.
		 *
		 * This is displayed in the code actions UI.
		 */
		reason: string;
	};
```

This PR shows the reason of the disabled action and stops the command execution.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -469,7 +469,8 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \           'codeActionKind': {
     \             'valueSet': ['', 'quickfix', 'refactor', 'refactor.extract', 'refactor.inline', 'refactor.rewrite', 'source', 'source.organizeImports'],
     \           }
-    \         }
+    \         },
+    \         'disabledSupport': v:true,
     \       },
     \       'codeLens': {
     \           'dynamicRegistration': v:false,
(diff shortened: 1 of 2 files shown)


hrsh7th:
Thank you for your contribution.

I think we should add `disabledSupport = true` for the ClientCapabilities.

rhysd:
Thank you for the point. I'll update the code.

rhysd:
I added `disabledSupport` capability to supported capabilities at b8d2a7d

hrsh7th:
LGTM!

hrsh7th (approved):

hrsh7th:
The code-action signs can be improved but it can be another PR, I think.

PR #1120, opened 2021-03-22
Support show message notification by handling `window/showMessage` response

rhysd:
This PR adds support for [ShowMessage Notification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_showMessage)

> The show message notification is sent from a server to a client to ask the client to display a particular message in the user interface.

This support is necessary to know the results of Code Lens with gopls.

I confirmed this PR worked fine on my local as follows:

tmp

(`"log"` is set to `g:lsp_show_message_log_level`)

TODO:

- [x] Add tests


mattn:
LGTM but pass to another look.

mattn:
I played with this and this works fine for me. So merging.

mattn:
Thank you

mattn:
@rhysd Ah, I forgot to check whether this change break (or not break) input(). Did you check this feature does not need to call inputsave/inputrestore?

rhysd:
@mattn I did not care about `inputsave`/`inputrestore`. As far as I tried this branch, I didn't see any problem on editing the text buffer while vim-lsp outputs server messages by `:echom`. But it might be safer to add `inputsave`/`inputrestore` before/after outputting a message.

PR #1121, opened 2021-03-22
Use `-wrap=0` in `*-nowrap` plug mappings

cuviper:
Using `--nowrap` doesn't work, but the related commands document that
you should use `-wrap=0`, which does work in the mappings too.


mattn:
Will merge after tests passed. Thank you

mattn:
Thank you

PR #1123, opened 2021-03-24
Use quickpick to choose Code Action

rhysd:
vim-lsp uses quickpick for Workspace Symbols and Code Lens, but uses `inputlist()` for Code Action.

`inputlist()` is not so useful because it always requires entering number. And using `inputlist()` only for Code Action is not consistent.

This PR changes Code Action to use quickpick to choose action.


mattn (approved):

mattn:
Works well.

mattn:
Thank you.

PR #1125, opened 2021-03-26
Silence plug mappings

obcat:
I added `<silent>`s to plug map's definitions to prevent flash of key sequences like `:call lsp#ui#vim#call_hierarchy_incoming()` in the command line.

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -135,46 +135,46 @@ command! LspDocumentFold call lsp#ui#vim#folding#fold(0)
 command! LspDocumentFoldSync call lsp#ui#vim#folding#fold(1)
 command! -nargs=? LspSemanticScopes call lsp#ui#vim#semantic#display_scope_tree(<args>)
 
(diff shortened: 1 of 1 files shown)


mattn:
I guess some echo commands should use unsilent.

obcat:
> I guess some echo commands should use unsilent.

I suppose there is no need to do that.

The ability of `map-<silent>` modifier is to silence corresponding key sequence showing in the cmdline. It does not silence messages from the executed commands:

```vimhelp
						*:map-<silent>* *:map-silent*
To define a mapping which will not be echoed on the command line, add
"<silent>" as the first argument.  Example: >
	:map <silent> ,h /Header<CR>
The search string will not be echoed when using this mapping.  Messages from
the executed command are still given though.  To shut them up too, add a
":silent" in the executed command: >
	:map <silent> ,h :exe ":silent normal /Header\r"<CR>
```

https://github.com/vim/vim/blob/f49a1fcdb952ca270f108fcc27bb8ad6922e0807/runtime/doc/map.txt#L201-L208

The result of test also proves that this PR does not introduce problems ;)

mattn:
Thank you

obcat:
Thank you too! 🎉

PR #1126, opened 2021-03-29
Select preferred action by default in quickpick by putting it at top of the results list

rhysd:
From [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction):

```typescript
export interface CodeAction {
	...

	/**
	 * Marks this as a preferred action. Preferred actions are used by the
	 * `auto fix` command and can be targeted by keybindings.
	 *
	 * A quick fix should be marked preferred if it properly addresses the
	 * underlying error. A refactoring should be marked preferred if it is the
	 * most reasonable choice of actions to take.
	 *
	 * @since 3.15.0
	 */
	isPreferred?: boolean;

	...
}
```

Since now quickpick is used to select Code Action results, an item at top of the results list can be run only typing `<CR>` quickly.

I thought putting a preferred action at top of the list is useful to run most reasonable action quickly.


hrsh7th:
LGTM. That's make sense.

PR #1127, opened 2021-03-31
Fix document layout and misspellings

rhysd:
This PR fixes document layout in `doc/vim-lsp.txt`. Only layout is tweaked and contents are not changed.

- Fix lines exceeding 78 characters
- Fix indent of code block is wrong
- Remove redundant blank lines
- Fix misspelled words


mattn:
Thanks

PR #1133, opened 2021-04-05
Show completion item detail when there is no documentation

tsufeki:
Currently we prepend `CompletionItem.detail` to popup when `documentation` exists, but when there's no `documentation` we show nothing even when `detail` exists. This PR fixes this.


mattn:
Thank you. Which language server did you test?

tsufeki:
@mattn ccls with c++: `detail` contains parent element (like class or namespace) but doesn't show in vim-lsp if item is not documented.

mattn:
Thank you

Issue #1107, opened 2021-03-16
too many 'workspace/didChangeConfiguration' may cause flood


zhu:
Currently `ensure_conf` which send `workspace/didChangeConfiguration`  is called in every `request_create`.
Some language server will do some extra works when `workspace/didChangeConfiguration` request.
For example, `pyls` will lint *all* documents in the workspace and send bunch of diagnostic notifications.

hrsh7th:
We should fix it. I will tackle it.

PR #1137, opened 2021-04-07
Send workspace_config only if changed

hrsh7th:
Fixes https://github.com/prabirshrestha/vim-lsp/issues/1107

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -616,12 +616,12 @@ endfunction
 function! s:ensure_conf(buf, server_name, cb) abort
     let l:server = s:servers[a:server_name]
     let l:server_info = l:server['server_info']
-    if has_key(l:server_info, 'workspace_config')
-        let l:workspace_config = l:server_info['workspace_config']
+    if has_key(l:server_info, 'workspace_config') && !get(l:server_info, '_workspace_config_sent', v:false)
+        let l:server_info['_workspace_config_sent'] = v:true
         call s:send_notification(a:server_name, {
             \ 'method': 'workspace/didChangeConfiguration',
             \ 'params': {
-            \   'settings': l:workspace_config,
+            \   'settings': l:server_info['workspace_config'],
             \ }
             \ })
     endif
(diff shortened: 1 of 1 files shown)


autoload/lsp.vim
```diff
@@ -616,12 +616,12 @@ endfunction
 function! s:ensure_conf(buf, server_name, cb) abort
     let l:server = s:servers[a:server_name]
     let l:server_info = l:server['server_info']
-    if has_key(l:server_info, 'workspace_config')
-        let l:workspace_config = l:server_info['workspace_config']
+    if has_key(l:server_info, 'workspace_config') && !get(l:server_info, '_workspace_config_sent', v:false)
```
rhysd, line 619: This means didChangeConfiguration is sent only once and never sent again. `lsp#update_workspace_config` no longer works. Is it ok?

hrsh7th: I think it should be work as expected because `lsp#update_workspace_config` will re-set flags.

rhysd: Hm, I didn't see how they are reset.

The implementation of `lsp#update_workspace_config` is as follows:

```vim
function! lsp#update_workspace_config(server_name, workspace_config) abort
    let l:server = s:servers[a:server_name]
    let l:server_info = l:server['server_info']
    if has_key(l:server_info, 'workspace_config')
        call s:merge_dict(l:server_info['workspace_config'], a:workspace_config)
    else
        let l:server_info['workspace_config'] = a:workspace_config
    endif
    call s:ensure_conf(bufnr('%'), a:server_name, function('s:Noop'))
endfunction
```

It merges new config into existing one and immediately calls `s:ensure_conf`. I couldn't find where the `_workspace_config_sent` flag is reset. Would you explain how the flag is reset?

rhysd: Ah, I overlooked https://github.com/prabirshrestha/vim-lsp/pull/1137/commits/c473b5c8cfdbc8cedcdb6c4ceaa746db61a74dca#diff-dd9e05ae74405827f8c7e93a94069e8386722ce20e643642aa37c4049f24b070R1198. Thanks for the explanation

rhysd (approved):

hrsh7th:
If someone reports any problem on this, I will tackle it.

PR #1144, opened 2021-04-13
Transform cygwin/msys2 paths to windows before encoding them to URIs

ElArtista:
Cygwin/MSYS2 paths have the form:
```
/c/Users/SomeUser/path
```
While native Windows paths have the form:
```
C:\Users\SomeUser\path
```

Currently vim-lsp transforms the paths to URIs to the following relevant forms:
Cygwin / MSYS2
```
file:///c/Users/SomeUser/path
```
Native Windows
```
file:///C:/Users/SomeUser/path
```

The Language Servers though that are commonly built and run in windows (e.g. eclipse jdt.ls) do recognize and function only using the second URI form, as this is a valid Windows file URI.

Current PR fixes this, by transforming paths when running vim in cygwin mode before encoding them.

Tested and makes eclipse jdt.ls functional with MSYS2 vim.

### DIFF
--- a/autoload/lsp/utils.vim
+++ b/autoload/lsp/utils.vim
@@ -72,7 +72,7 @@ function! s:encode_uri(path, start_pos_encode, default_prefix) abort
 endfunction
 
 let s:path_to_uri_cache = {}
-if has('win32') || has('win64')
+if has('win32') || has('win64') || has('win32unix')
     function! lsp#utils#path_to_uri(path) abort
         if has_key(s:path_to_uri_cache, a:path)
             return s:path_to_uri_cache[a:path]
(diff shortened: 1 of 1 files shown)


autoload/lsp/utils.vim
```diff
@@ -82,16 +82,22 @@ if has('win32') || has('win64')
             let s:path_to_uri_cache[a:path] = a:path
             return s:path_to_uri_cache[a:path]
         else
+            " Transform cygwin paths to windows paths
+            let l:path = a:path
+            if has('win32unix')
+                let l:path = substitute(a:path, '\c^/\([a-z]\)', '\U\1:', '')
```
mattn, line 88: This break `substitute('/path/to/dir', '\c^/\([a-z]\)', '\U\1:', '')`.

mattn: Probably this is correct.
```
let l:path = substitute(l:path, '\c^/\([a-z]\)/', '\U\1:/', '')
```

ElArtista: Yes you are correct. Resolved

ElArtista:
Seems that the reverse operation must also be performed, converting paths to uris for msys2/cygwin too, making some additional changes for this to be complete

ElArtista:
Reverse conversion implemented, PR is now ready

autoload/lsp/utils.vim
```diff
@@ -117,13 +117,20 @@ else
...
         if has_key(s:uri_to_path_cache, a:uri)
             return s:uri_to_path_cache[a:uri]
         endif
 
-        let s:uri_to_path_cache[a:uri] = substitute(s:decode_uri(a:uri[len('file:///'):]), '/', '\\', 'g')
+        let l:path = substitute(s:decode_uri(a:uri[len('file:///'):]), '/', '\\', 'g')
+
+        " Transform windows paths to cygwin paths
+        if has('win32unix')
+            let l:path = substitute(l:path, '\c^\([A-Z]\):/', '/\l\1/', '')
```
mattn, line 130: The last slash (`'\c^\([A-Z]\):/'`) are already replaced with backslashes in above. So this should not be matched, I think.

ElArtista: You are correct, fixed this too along with the transformation from backslashes to slashes too

mattn:
Thank you

PR #1146, opened 2021-04-17
Make g:lsp_auto_enable work when lazily loaded

Bakudankun:
When vim-lsp is loaded after vim's starting (e.g. with `:packadd`), it is not enabled regardless of the value of `g:lsp_auto_enable`. So if you want to *lazily* load vim-lsp, you need to call `lsp#enable()` manually although the variable exists.
This PR makes vim-lsp to call `lsp#enable()` if `g:lsp_auto_enable` is TRUE in such condition. I think this behavior is what users expect from the variable's name.

This PR have a side effect to change the default behavior written above, because `g:lsp_auto_enable` is TRUE by default. You can disable its auto-enabling by setting `g:lsp_auto_enable = 0`.
I think most users who want to load vim-lsp on demand also want to enable vim-lsp at the same time, so this will not be a big problem.

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -69,13 +69,6 @@ let g:lsp_work_done_progress_enabled = get(g:, 'lsp_work_done_progress_enabled',
 
 let g:lsp_get_supported_capabilities = get(g:, 'lsp_get_supported_capabilities', [function('lsp#default_get_supported_capabilities')])
 
-if g:lsp_auto_enable
-    augroup lsp_auto_enable
-        autocmd!
-        autocmd VimEnter * call lsp#enable()
-    augroup END
-endif
-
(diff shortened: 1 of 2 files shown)


mattn:
Tests failing

Bakudankun:
Tests failed because they relied on the previous behavior that `lsp#enable()` is not called automatically after VimEnter.
Adding `let g:lsp_auto_enable = 0` to the `.themisrc` resolved them.

Bakudankun:
> I think most users who want to load vim-lsp on demand also want to enable vim-lsp at the same time, so this will not be a big problem.

Now I'm thinking this may not be true...
With the new default behavior, `lsp#register_server()` must be called in `lsp_setup` autocmd because `lsp#enable()` is called right after `lsp#register_server()` become available.
Hmm...

PR #1190, opened 2021-06-08
diagnostics list can be v:null

mattn:
Ref #1186

### DIFF
--- a/autoload/lsp/internal/diagnostics/first_line.vim
+++ b/autoload/lsp/internal/diagnostics/first_line.vim
@@ -15,10 +15,14 @@ function! lsp#internal#diagnostics#first_line#get_first_error_line(options) abor
 
     let l:first_error_line = v:null
     for l:diagnostics_response in values(l:diagnostics_by_server)
-        for l:diagnostics in l:diagnostics_response['params']['diagnostics']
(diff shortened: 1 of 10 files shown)


rhysd:
I just grepped around the repository and found more places which might need to be fixed:

- https://github.com/prabirshrestha/vim-lsp/blob/61916a2ddd454308cc6f8413d6b8becc5c956917/autoload/lsp/internal/diagnostics/movement.vim#L174
- https://github.com/prabirshrestha/vim-lsp/blob/61916a2ddd454308cc6f8413d6b8becc5c956917/autoload/lsp/internal/diagnostics/movement.vim#L178
- https://github.com/prabirshrestha/vim-lsp/blob/61916a2ddd454308cc6f8413d6b8becc5c956917/autoload/lsp/internal/diagnostics/state.vim#L166

Would you check them?

rhysd (changes requested):
https://github.com/prabirshrestha/vim-lsp/pull/1190#issuecomment-856904226

mattn:
@rhysd Thank you. Fixed

rhysd (approved):
Thank you for the fixes. Looks good.

mattn:
Merging. I will fix if still have same issue.

PR #1202, opened 2021-07-20
Silence initialization message of documentation buffer

itchyny:
Recently I see `"VS.Vim.Buffer: 1: VS.Vim.Buffer.Default" [New]` message on initialization of this plugin. This is caused by [`bufload` function here](https://github.com/prabirshrestha/vim-lsp/blob/fb0a72306e1fc42c1ce8909aa1e420ddfccf6fbf/autoload/vital/_lsp/VS/Vim/Buffer.vim#L67), and was updated dropping `silent` by [this commit](https://github.com/hrsh7th/vim-vital-vs/commit/5a5c69d67777c88adc7c47cf0a24b2207d9c83f3#diff-2ef35f69ca1fd03b90597a9f00e91d07377a63e132fd15fc6b3e6f2bc8721180L44). We don't need information vim-lsp opens a new buffer.
ping: #1178


hrsh7th:
LGTM.

PR #1205, opened 2021-08-02
Update vital-vs

hrsh7th:
This PR aims to solve the wrong augroup creation.

The older implementation wouldn't create correct augroup (it will be `printf('VS_Vim_Window_FloatingWindow:...)`.
The newer implementation will create expected augroup that will be filepath instead.


hrsh7th:
The test fails seem to cause by latest neovim.

hrsh7th:
I've fixed for latest vim/neovim.

prabirshrestha (approved):

PR #1225, opened 2021-09-23
Fix completion sort

rhysd:
This fixes implementation by #1223.

Vim script's string is always falsy:

```
:echo !!'foo'
0
```

So, `x ? 1 : 2` is always `1` when `x` is string. `empty()` should be used to check emptiness.

CC: @heavenshell

### DIFF
--- a/test/lsp/omni.vimspec
+++ b/test/lsp/omni.vimspec
@@ -424,5 +424,65 @@ Describe lsp#omni
             \}
             Assert Equals(lsp#omni#get_vim_completion_items(options), want)
         End
+
+        Describe g:lsp_ignorecase
+            Before all
+                let saved_ignorecase = get(g:, 'lsp_ignorecase', v:null)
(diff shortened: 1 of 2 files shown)


test/lsp/omni.vimspec
```diff
@@ -185,19 +185,19 @@ Describe lsp#omni
...
             \ 'kind': '3',
-            \ 'sortText': '1'
+            \ 'sortText': 'a'
             \},
             \{
-            \ 'label': 'my-label2',
+            \ 'label': 'my-label3',
             \ 'kind': '3',
-            \ 'sortText': '2'
+            \ 'sortText': 'b'
```
rhysd, line 200: Previously these tests passed because number strings like `'1'` are implicitly converted to numbers like `1`.

heavenshell:
Oops, Thank you for fixing!

PR #1226, opened 2021-09-24
Make copen configurable

impactaky:
I would like to be able to customize the way qflist opens.


mattn:
Nitpick. this. `g:lsp_copen_command` is not set in anywhere.

impactaky:
Thank you! I fixed it.

PR #1228, opened 2021-09-28
Avoid BufEnter autocmd when opening floating window

hrsh7th:
The `s:Buffer.create()` will fire some of the autocmd like `FileType` `BufEnter` etc... due to `bufload` calls.

1. The `bufload` is needed to call `setbufline` before showing the buffer.
2. Basically, `FileType` autocmd is needs for syntax highlighting.

So in `vsim-vital-vs` context, `s:Buffer.create()` should fire autocmd but in this case we don't need `FileType` autocmd because syntax highlighting will be applied by calling `s:Markdown.apply`.

So we can avoid autocmd here.


rhysd (approved):

hrsh7th:
Thank you for review!

PR #1229, opened 2021-09-28
Fix Neovim CI

rhysd:
Now Neovim nightly packages are missing due to [job error](https://github.com/neovim/neovim/actions/runs/1282732454)


rhysd:
I confirmed CI passed even if Neovim nightly CI workflow failed.

https://github.com/prabirshrestha/vim-lsp/pull/1229/checks?check_run_id=3733481186

PR #1235, opened 2021-11-02
Update CI dependencies; Go to 1.17 and gopls to 0.7.3

rhysd:


rhysd:
will `git push -f` to check if a cache works

rhysd:
It works. Merging.

PR #1246, opened 2021-12-24
Fix hover preview height

hrsh7th:
The `l:contents` is group of contents so we should use `l:lines` as previewheight.

### DIFF
--- a/autoload/lsp/internal/document_hover/under_cursor.vim
+++ b/autoload/lsp/internal/document_hover/under_cursor.vim
@@ -101,18 +101,19 @@ function! s:show_preview_window(server_name, request, response) abort
         return
     endif
 
+    let l:lines = lsp#utils#_split_by_eol(join(l:contents, "\n\n"))
     let l:view = winsaveview()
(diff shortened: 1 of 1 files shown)


autoload/lsp/internal/document_hover/under_cursor.vim
```diff
@@ -101,18 +101,19 @@ function! s:show_preview_window(server_name, request, response) abort
         return
     endif
 
+    let l:lines = lsp#utils#_split_by_eol(join(l:contents, "\n\n"))
```
mattn, line 104: What do you mean this "\n\n" ?

hrsh7th: The `l:contents` is meant the `multiple grouped contents`. So I think the `\n\n` is needed to separate content groups.

mattn: Ah, make sense.

mattn (approved):

PR #1248, opened 2021-12-28
Fix conceallevel=2 to LspHover preview

heavenshell:
Hi, I'm using `g:lsp_hover_ui = 'preview'`.
When hover with preview window, conceal has been shown.

Before
スクリーンショット 2021-12-29 1 07 33

After
スクリーンショット 2021-12-29 1 08 41

When using `float`, concellevel set as `2`.
https://github.com/prabirshrestha/vim-lsp/blob/858a5adeaae6283b8a061c8604b875be7048b2e3/autoload/lsp/internal/document_hover/under_cursor.vim#L233

I add same value to preview.
But, should we use `g:lsp_hover_conceal` value instead?


mattn (approved):

hrsh7th:
LGTM!

heavenshell:
Thank you!

PR #1253, opened 2022-01-02
Fix hover position

hrsh7th:
Fix #1245

PR #1254, opened 2022-01-02
Handle error case where cursor is outside screen

rhysd:
This is small followup of #1253. `screenpos()` returns `{'col': 0, 'row': 0, 'endcol': 0, 'curscol': 0}` when the position is outside screen. This patch adds error handling for the case.


hrsh7th (approved):
LGTM

PR #1279, opened 2022-02-24
Print workspace_config with :verbose LspStatus

idbrii:
When trying to debug server configuration (recently for me: mattn/vim-lsp-settings/issues/542), it's useful to see the workspace_config for the server. I think this is more useful than anything else in server_info which are more session-specific (files).

This PR allows you to use ` :verbose LspStatus` to add the workspace_config to the output. I uses tpope/vim-scriptease's pretty printing if available (new lines and indents) and otherwise falls back to json encoding (one big line).

I can squash the changes if you'd like. Or change it from verbose to bang.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,13 @@ let g:lsp_log_file = expand('~/vim-lsp.log')
 let g:asyncomplete_log_file = expand('~/asyncomplete.log')
 ```
 
+You can get detailed status on your servers using `:CheckHealth` -- built into neovim or in a plugin on vim:
+
+```vim
+if !has('nvim') | Plug 'rhysd/vim-healthcheck' | endif
+CheckHealth
+```
+
 ## Tests
 
 [vim-themis](https://github.com/thinca/vim-themis) is used for testing. To run
--- a/autoload/health/lsp.vim
+++ b/autoload/health/lsp.vim
@@ -30,11 +30,11 @@ function! health#lsp#check() abort
         call health#report_warn('no servers connected')
     endif
 
-    call health#report_start("server configuration")
     for l:k in sort(keys(l:server_status))
+        call health#report_start(printf('server configuration: %s', l:k))
         let l:report = l:server_status[l:k]
 
-        let l:msg = printf("**%s** configuration:\n", l:k)
+        let l:msg = "\t\n"
         let l:msg .= s:BuildConfigBlock('allowlist', l:report.info)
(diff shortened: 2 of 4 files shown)


prabirshrestha:
Can you past a screenshot of how the output looks like?

I have been thinking about use `:CheckHealth` as an addition to verbose `:LspStatus` as `CheckHealth` is bundled with neovim and there is a polyfil https://github.com/rhysd/vim-healthcheck. 

http://vimcasts.org/episodes/neovim-checkhealth/

prabirshrestha:
Meanwhile I have filed a feature request to vim to bundle `:CheckHealth` by default. https://github.com/vim/vim/issues/9838

idbrii:
Lua was the specific server I was trying to configure, but here's two other servers I use (where I use the default config):

with scriptease
with scriptease
and without
with jsonencode

For CheckHealth, you mean implement `health#lsp#check()`? From [the docs](https://neovim.io/doc/user/pi_health.html), it looks like we'd use `health#report_info()` since we don't actually do any error detection here. But if we had other validation (server doesn't exist for current filetype), we could print some warnings or errors.

prabirshrestha:
Yes we should implement `health#lsp#check()`. Problem with adding verbose log with message is that there could be lot of servers running and users will have to scroll.

Another option is to continue with `:LspStatus <servername>` that prints to buffer with information but then this would similar to the healthcheck. The reason I was preferring `:checkhealth` is so that we can print information about features enabled/disable which will make it easy for anyone to discover other potential issues.

idbrii:
I implemented checkhealth support:

image

I also included advice when status is "failed":
image


Do you want me to remove the verbose log with message? checkhealth strips out leading indentation so the workspace output doesn't look as nice, but I agree that it's better for it to get dumped to a buffer.

prabirshrestha:
This looks great. Seems like there are some lint issues that needs to be fixed. Also would be good to document `:Healthcheck`. Can also update [debugging section](https://github.com/prabirshrestha/vim-lsp#debugging).

I would also add this for vim folks.

```vim
if !has('nvim') | Plug 'rhysd/vim-healthcheck' | endif
```

idbrii:
Added documentation and included a Plug reference in the readme, but not in code since vim-plug is not mandatory. Looks like all lint checks pass now.

prabirshrestha:
Merged. Thanks

Issue #1300, opened 2022-04-17
Completions with `textEdit` suggestions break autocomplete


EoinDavey:
Consider the following behaviour for the `typescript-language-server` LSP server.

If you have a variable `s` of type `string` and you type `s.|` (where `|` is the cursor position), activating omni-complete *should* suggest string methods e.g. `s.includes`, `s.trimStart` etc. However only one suggestion is returned: `s[Symbol]`.

What is happening seems to be that, as the `[Symbol]` suggestion needs to delete the already typed `.`, it returns a `textEdit` completion with the start character set to the character _before the `.`_.

As of #1236 this format is supported, but the implementation takes whatever the minimum starting character is across all completion suggestions. This breaks the other completion suggestions which are not intended to be inserted after the `.`, not replacing it.

They're broken in two ways:
1. The `last_typed_word` in the `omni.vim` script matches `.`, rather than the empty string as it would have before if the start column had not moved. This means that the `prefix` and `contains` filtering of completions exclude all the method completions as they have labels with no `.` character e.g. "at", "includes" etc.
2. Even if the filtering is avoided (e.g. by setting `none` filter) then the insertions happen at the wrong location. i.e. `s.|` with the completion `includes` should complete to `s.includes`, but instead it completes to `sincludes`.

None of this happens if `[Symbol]` is not in the returned completions, as the minimum start column stays at the correct location.

A fix for this could be to find the minimum start character, then for any words that are not supposed to start at this position, add the missing prefix to the start of these completions. e.g. in the above case turn "includes", "at", "charAt" etc to ".includes", ".at" and ".charAt".

EoinDavey:
I've created a pull request with a fix for this, but I'm not very familiar with vimscript so it likely needs major changes

https://github.com/prabirshrestha/vim-lsp/pull/1301

EoinDavey:
@hrsh7th @mattn keeping this issue open, question on https://github.com/prabirshrestha/vim-lsp/pull/1301 still stands

PR #1301, opened 2022-04-17
Add prefix to completions if start col changes.

EoinDavey:
See issue: https://github.com/prabirshrestha/vim-lsp/issues/1300

`textEdit` completions can cause the column where the completion begins
to change. Current behaviour is to just take the minimum of such start
characters. This breaks autocompletion of edits that don't start at this
earliest position as:

(a) They are filtered out by `prefix` or `contains` autocomplete filters
as they no longer match the `last_typed_word`
(b) They insert at the wrong location even if the above is fixed.

The solution included here is to add any extra characters included by
the moving of the start column as a prefix to any insertion completions.

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -340,7 +340,7 @@ function! lsp#omni#get_vim_completion_items(options) abort
             let l:character = l:completion_item['textEdit']['range']['start']['character']
             let l:vim_complete_item['word'] = l:cur_line[l:start_character : l:character - 1] . l:completion_item['textEdit']['newText']
         elseif has_key(l:completion_item, 'insertText') && !empty(l:completion_item['insertText'])
-            let l:with_prefix = l:cur_line[l:start_character:s:completion['startcol'] - 2] . l:completion_item['insertText']
+            let l:with_prefix = l:cur_line[l:start_character : s:completion['startcol'] - 2] . l:completion_item['insertText']
             let l:completion_item['insertText'] = l:with_prefix
             let l:vim_complete_item['word'] = l:with_prefix
         else
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -319,6 +319,13 @@ function! lsp#omni#get_vim_completion_items(options) abort
     endif
 
     let l:start_character = l:complete_position['character']
+
+    for l:completion_item in l:items
+        if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && has_key(l:completion_item['textEdit'], 'range') && has_key(l:completion_item['textEdit'], 'newText')
+            let l:start_character = min([l:completion_item['textEdit']['range']['start']['character'], l:start_character])
+        endif
+    endfor
+
     let l:vim_complete_items = []
     for l:completion_item in l:items
         let l:expandable = get(l:completion_item, 'insertTextFormat', 1) == 2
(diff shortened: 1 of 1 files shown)


hrsh7th:
It's correct approach, I think Thank you. I'll review it.

autoload/lsp/omni.vim
```diff
@@ -329,12 +336,12 @@ function! lsp#omni#get_vim_completion_items(options) abort
             \ 'icase': 1,
             \ }
         if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && has_key(l:completion_item['textEdit'], 'range') && has_key(l:completion_item['textEdit'], 'newText')
-            let l:vim_complete_item['word'] = l:completion_item['textEdit']['newText']
-            let l:start_character = min([l:completion_item['textEdit']['range']['start']['character'], l:start_character])
+            let l:character = l:completion_item['textEdit']['range']['start']['character']
+            let l:vim_complete_item['word'] = getline('.')[l:start_character:l:character - 1] . l:completion_item['textEdit']['newText']
```
hrsh7th, line 340: Could you create new variable for `getline('.')`?

hrsh7th:
LGTM. I'll merge this after solve the reviewed point.

autoload/lsp/omni.vim
EoinDavey: Done

EoinDavey:
Hmm there seems to be something wrong with this, but I'm not sure what exactly. Don't merget as-is

EoinDavey:
Ok, it turns out that the raw response data also needs to have the prefix appended to it as it gets used by functions called by `CompleteDone` event to overwrite the inserted text in autoload/lsp/ui/vim/completion.vim. I don't fully understand what that script does but the previous commit addresses the issue

autoload/lsp/omni.vim
```diff
@@ -329,12 +337,16 @@ function! lsp#omni#get_vim_completion_items(options) abort
             \ 'icase': 1,
             \ }
         if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && has_key(l:completion_item['textEdit'], 'range') && has_key(l:completion_item['textEdit'], 'newText')
-            let l:vim_complete_item['word'] = l:completion_item['textEdit']['newText']
-            let l:start_character = min([l:completion_item['textEdit']['range']['start']['character'], l:start_character])
+            let l:character = l:completion_item['textEdit']['range']['start']['character']
+            let l:vim_complete_item['word'] = l:cur_line[l:start_character:l:character - 1] . l:completion_item['textEdit']['newText']
```
mattn, line 341:
```suggestion
            let l:vim_complete_item['word'] = l:cur_line[l:start_character : l:character - 1] . l:completion_item['textEdit']['newText']
```

autoload/lsp/omni.vim
```diff
@@ -329,12 +337,16 @@ function! lsp#omni#get_vim_completion_items(options) abort
...
         elseif has_key(l:completion_item, 'insertText') && !empty(l:completion_item['insertText'])
-            let l:vim_complete_item['word'] = l:completion_item['insertText']
+            let l:with_prefix = l:cur_line[l:start_character:s:completion['startcol'] - 2] . l:completion_item['insertText']
```
mattn, line 343:
```suggestion
            let l:with_prefix = l:cur_line[l:start_character : s:completion['startcol'] - 2] . l:completion_item['insertText']
```

autoload/lsp/omni.vim
```diff
@@ -329,12 +337,16 @@ function! lsp#omni#get_vim_completion_items(options) abort
...
+            let l:completion_item['insertText'] = l:with_prefix
+            let l:vim_complete_item['word'] = l:with_prefix
         else
-            let l:vim_complete_item['word'] = l:completion_item['label']
+            let l:with_prefix = l:cur_line[l:start_character:s:completion['startcol'] - 2] . l:completion_item['label']
```
mattn, line 347:
```suggestion
            let l:with_prefix = l:cur_line[l:start_character : s:completion['startcol'] - 2] . l:completion_item['label']
```

mattn:
LGTM.

mattn:
Test seems failing.

EoinDavey:
> Test seems to be failing.

Yeah, it seems to be an issue where `s:completion['startcol']` isn't defined in the unit tests because it's a global variable. I don't know whether to just instantiate the global variable in the test or not, as the problem seems more rooted in the fact that relying on the global variable like this isn't good practice, and we should be taking in the column where the completion will start as a parameter.

What do you think?

PR #1311, opened 2022-05-09
Fix `tagfunc` name as documented

satorunooshie:
https://github.com/prabirshrestha/vim-lsp/blob/master/doc/vim-lsp.txt#L1879-L1885
https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/tag.vim#L144


mattn:
In Vim script, filename of .vim become namespace. So code is right. This is documentation bug.

satorunooshie:
Please ignore this PR because it was my misunderstanding.
https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp.vim#L1153-L1155

PR #1313, opened 2022-05-13
Improve tag location string

whenov:
Old string: \%10l\%5c
New string: 10L, 5C

In the old string, the escape character is reduntant, the % character is cumbersome, and the lower case "l" looks like "1".


mattn:
cmd should be Ex command to find the tag. AFAIK, it seems not be an Ex command.

PR #1321, opened 2022-06-06
Fix when textEdit type was InsertReplaceEdit

wordijp:
When using fuzzy match, when lsp returns InsertReplaceEdit, range was not found error.
(I used volar-server when writing Vue)

1. setup fuzzy match for asyncomplete.vim in vimrc

```vim
" vimrc
function! s:fuzzy_preprocessor(options, matches) abort
  let l:base = a:options["base"]

  " NOTE: There is an lsp with a '.' At the beginning when it is a member candidate.
  "       ex) volar-server
  if len(l:base) >= 2 && l:base[0] == '.'
    let l:base = l:base[1:]
    let a:options["startcol"] += 1
  endif

  let l:items = []
  for l:matches in values(a:matches)
    if len(l:base) > 0
      let l:items += filter(copy(l:matches['items']), '!empty(matchfuzzy([v:val["word"]], l:base))')
    else
      let l:items += l:matches['items']
    endif
  endfor

  call asyncomplete#preprocess_complete(a:options, l:items)
endfunction

let g:asyncomplete_preprocessor = [function('s:fuzzy_preprocessor')]
```

2. try fuzzy match for vue code.

```vue
<script lang="ts" setup>
const a = [];
a.flt|
     ^
  try fuzzy match expand `filter`
</script>
```

3. result completionItem json

```json
{
  "label": "filter",
  "commitCharacters": [
    ".",
    ",",
    ";",
    "("
  ],
  "data": {
    "uri": "file:///e%3A/work/Web/vue/vue-typescript/src/App.vue.ts",
    "fileName": "e:/work/Web/vue/vue-typescript/src/App.vue.ts",
    "offset": 56,
    "originalItem": {
      "name": "filter",
      "kindModifiers": "declare",
      "sortText": "11",
      "kind": "method"
    }
  },
  "textEdit": {
    "replace": {
      "end": {
        "character": 5,
        "line": 2
      },
      "start": {
        "character": 2,
        "line": 2
      }
    },
    "newText": "filter",
    "insert": {
      "end": {
        "character": 5,
        "line": 2
      },
      "start": {
        "character": 2,
        "line": 2
      }
    }
  },
  "sortText": "11",
  "kind": 2,
  "detail": "(method) Array<any>.filter<any>(predicate: (value: any, index: number, array: any[]) => value is any, thisArg?: any): any[] (+1 overload)",
  "documentation": {
    "kind": "markdown",
    "value": "Returns the elements of an array that meet the condition specified in a callback function.\n\n*@param* `predicate` — A function that accepts up to three arguments. The filter method calls the predicate function one time for each element in the array.  \n\n*@param* `thisArg` — An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value."
  },
  "insertTextFormat": 1
}
```

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -328,9 +328,10 @@ function! lsp#omni#get_vim_completion_items(options) abort
             \ 'empty': 1,
             \ 'icase': 1,
             \ }
-        if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && has_key(l:completion_item['textEdit'], 'range') && has_key(l:completion_item['textEdit'], 'newText')
+        if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && has_key(l:completion_item['textEdit'], 'newText')
+            let l:text_edit_range = lsp#utils#text_edit#_get_range_maybe_replace(l:completion_item['textEdit'])
             let l:vim_complete_item['word'] = l:completion_item['textEdit']['newText']
-            let l:start_character = min([l:completion_item['textEdit']['range']['start']['character'], l:start_character])
+            let l:start_character = min([l:text_edit_range['start']['character'], l:start_character])
         elseif has_key(l:completion_item, 'insertText') && !empty(l:completion_item['insertText'])
             let l:vim_complete_item['word'] = l:completion_item['insertText']
         else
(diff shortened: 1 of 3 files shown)


wordijp:
According to this issue, the choice of insert or replace is left to the client, so it is a protocol that is still starting to appear, so it was supported by replace, I think that the option to select insert is necessary in the future.

microsoft/language-server-protocol#1260

mattn:
Thank you. This problem is fixed by https://github.com/prabirshrestha/vim-lsp/pull/1340

wordijp:
OK, Thank you.

PR #1334, opened 2022-07-16
Fix completion documentation floating window

yuilib:
With the current implementation, if there is no enough space to show a floating window for completion documentation on both the right and left sides, the floating window simply doesn't show.
This patch fixed the issue: When a margin on the right side is not enough to show a floating window, the window is shown on the side which has the bigger margin.


mattn:
Looks good to me.

mattn:
Thank you

yuilib:
#1346 should fix the issue.

yarko3:
Great; thanks for the fix!

PR #1340, opened 2022-07-21
Implemented word correction in cases where startcol was corrected by …

hrsh7th:
…TextEdit.range.start.character

# Problem

In typescript-language-server, the vim-lsp's completion mechanism can't handle the class static property completion.

# Solution

Implement `complete_item.word` fixing.

# Background

Currently, vim-lsp is fixing the `startcol` by `TextEdit.range.start.character`.
But it is not enough. We should fix `complete_item.word` too.

For example, in the following cases, word correction processing is required.

1. The line is `Class.|`
2. The `CompletionItem`
    - `{ label: 'PROP' }
    - `{ label: 'Symbol', textEdit: { newText: '[Symbol]' ... } }`

In such case, the `startcol` should be the `Class|.`.
This is because the `[Symbol]` item expects a `Class[Symbol]`.

 So we should add `.` for all item in this case.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -95,7 +95,7 @@ function! lsp#get_server_names() abort
 endfunction
 
 function! lsp#get_server_info(server_name) abort
-    return s:servers[a:server_name]['server_info']
+    return get(get(s:servers, a:server_name, {}), 'server_info', {})
 endfunction
 
 function! lsp#get_server_root_uri(server_name) abort
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -358,7 +358,6 @@ function! lsp#omni#get_vim_completion_items(options) abort
             if l:start_character < l:item_start_character
                 let l:item = l:vim_complete_items[l:i]
                 let l:item['word'] = strcharpart(l:current_line, l:start_character, l:item_start_character - l:start_character) .. l:item['word']
-                let l:item['abbr'] = l:item['word']
             endif
         endfor
     endif
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -325,7 +325,7 @@ function! lsp#omni#get_vim_completion_items(options) abort
             let l:item_start_character = l:completion_item['textEdit']['range']['start']['character']
             if l:item_start_character < l:default_start_character
                 " Add already typed word. The typescript-language-server returns `[Symbol]` item for the line of `Hoo.|`. So we should add `.` (`.[Symbol]`) .
-                let l:vim_complete_item['word'] = strcharpart(l:current_line, l:item_start_character, l:default_start_character - l:item_start_character) .. l:vim_complete_item['word']
+                let l:vim_complete_item['word'] = strcharpart(l:current_line, l:item_start_character, l:default_start_character - l:item_start_character) . l:vim_complete_item['word']
             endif
             let l:start_character = min([l:item_start_character, l:start_character])
             let l:start_characters += [l:item_start_character]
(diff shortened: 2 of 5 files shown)


autoload/lsp/omni.vim
```diff
@@ -59,23 +59,10 @@ function! lsp#omni#complete(findstart, base) abort
...
-        let l:curpos = getcurpos()
-        let l:left = strpart(getline(l:curpos[1]), 0, l:curpos[2]-1)
+        let l:left = strpart(getline('.'), 0, col('.')-1)
 
         " Initialize the default startcol. It will be updated if the completion items has textEdit.
-        let s:completion['startcol'] = 1 + matchstrpos(l:left, l:refresh_pattern)[1]
-        if s:completion['startcol'] == 0
-            let s:completion['startcol'] = strlen(l:left) + 1
-        endif
+        let s:completion['startcol'] = s:get_startcol(l:left, l:info['server_names'])
```
hrsh7th, line 65: It's just refactor. The logic is moved to `s:get_startcol`.

autoload/lsp/omni.vim
```diff
@@ -298,6 +285,9 @@ function! lsp#omni#get_vim_completion_items(options) abort
     let l:server_name = l:server['name']
     let l:kind_text_mappings = s:get_kind_text_mappings(l:server)
     let l:complete_position = a:options['position']
+    let l:current_line = getline('.')
+    let l:default_startcol = s:get_startcol(strcharpart(l:current_line, 0, l:complete_position['character']), [l:server_name])
+    let l:default_start_character = strchars(strpart(l:current_line, 0, l:default_startcol - 1))
```
hrsh7th, line 290: getting `startcol` that considered the `refresh_pattern`.

autoload/lsp.vim
```diff
@@ -95,7 +95,7 @@ function! lsp#get_server_names() abort
 endfunction
 
 function! lsp#get_server_info(server_name) abort
-    return s:servers[a:server_name]['server_info']
+    return get(get(s:servers, a:server_name, {}), 'server_info', {})
```
hrsh7th, line 98: This is needed to pass test.

New implementation uses `get_server_info` in `lsp#omni#get_vim_completion_items` so I needed this...

autoload/lsp/omni.vim
```diff
@@ -95,17 +82,6 @@ function! lsp#omni#complete(findstart, base) abort
...
-        let l:filter_text = l:user_data['completion_item']['filterText']
-        if empty(l:filter_text) && has_key(l:user_data['completion_item'], 'label')
-            " When filterText is `falsy` the label is used as the filter text
-            let l:filter_text = l:user_data['completion_item']['label']
-        endif
-        if !empty(l:filter_text)
-            return lsp#utils#_trim(l:filter_text)
-        endif
-    endif
     return lsp#utils#_trim(a:item['word'])
```
hrsh7th, line 85: This PR fixes the `v:completed_item['word']` but this method uses `CompletionItem.filterText`.
So I think we should use `v:completed_item['word']` instead. (This behavior is the same as the asyncomplete-lsp.vim's one).

autoload/lsp/omni.vim
```diff
@@ -369,12 +372,13 @@ endfunction
...
 "
-function! s:create_user_data(completion_item, server_name, complete_position) abort
+function! s:create_user_data(completion_item, server_name, complete_position, start_character) abort
     let l:user_data_key = s:create_user_data_key(s:managed_user_data_key_base)
     let s:managed_user_data_map[l:user_data_key] = {
     \   'complete_position': a:complete_position,
     \   'server_name': a:server_name,
-    \   'completion_item': a:completion_item
+    \   'completion_item': a:completion_item,
+    \   'start_character': a:start_character,
```
hrsh7th, line 381: Save the start_character as it may be different for each item.

autoload/lsp/utils/text_edit.vim
```diff
@@ -1,3 +1,11 @@
+function! lsp#utils#text_edit#get_range(text_edit) abort
+  let l:insert = get(a:text_edit, 'insert', v:null)
+  if type(l:insert) == v:t_dict
+    return l:insert
+  endif
+  return get(a:text_edit, 'range', v:null)
+endfunction
```
hrsh7th, line 7: Minimamu `insertReplaceEdit` support.

mattn:
Works fine for me.

hrsh7th:
html/css/json also works fine.

mattn (approved):

PR #1358, opened 2022-09-18
Add support for virtual text diagnostics in Vim9

dotdash:


prabirshrestha:
could you also add screenshots?

dotdash:
ss2
ss1

dotdash:
The `» ` prefix is from `let g:lsp_diagnostics_virtual_text_prefix = '» '`

dotdash:
Ah, I see vim-lsp only delays the processing of incoming diagnostics, but then processes all of them, including outdated ones, and the textprop API then complains if a diagnostics refers to a line that no longer exists. I'll have to take a closer look at how things work, and decide whether the processing can be improved in general, or if I'll just have to drop diagnostics that refer to lines that no longer exist.

Thanks!

prabirshrestha:
This is async problem, you can skip diagnostics that are in invalid line or column. We probably want `is_valid_range()` method that we can user everywhere for future async problems.  Having flags such as disabling diagnostics for current line or insert mode can also help.

astronautindisguise:
Any progress on this?

dotdash:
Should be good to go

prabirshrestha:
Merged. Thanks. This looks great been using it and loving it!

PR #1373, opened 2022-10-14
Allow `workspace_config` to be a function

ilya-bobyr:
It is sometimes useful to generate `workspace_config` using a callback. For example, if the produced config needs to include some information from the actual workspace.


ilya-bobyr:
Closed, in favor of https://github.com/prabirshrestha/vim-lsp/pull/1400

PR #1379, opened 2022-10-17
Fix default supported capabilities

tsukkee:
This PR adds some capabilities which are referred from https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/capabilities.vim but are not set at `lsp#default_get_supported_capabilities`, then sorts them alphabetically again.

I assume that these capabilities might be expected to be registered dynamically so these are missing in `lsp#default_get_supported_capabilities`. But, some language servers do not support dynamic registration and this means they will not recognize these capabilities are supported by vim-lsp.

So, I'm wondering whether `'dynamicRegistration'` property of these capabilities should be `v:true` or `v:false`. 
It would be appreciate if giving advice here.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -485,7 +485,7 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     return {
     \   'textDocument': {
     \       'callHierarchy': {
-    \           'dynamicRegistration': v:true,
+    \           'dynamicRegistration': v:false,
     \       },
     \       'codeAction': {
     \         'dynamicRegistration': v:false,
@@ -561,7 +561,7 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \           'dynamicRegistration': v:false,
     \       },
     \       'rename': {
-    \           'dynamicRegistration': v:true,
+    \           'dynamicRegistration': v:false,
     \           'prepareSupport': v:true
     \       },
     \       'semanticTokens': {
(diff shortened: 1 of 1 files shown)


autoload/lsp.vim
```diff
@@ -484,6 +484,9 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     " Sorted alphabetically
     return {
     \   'textDocument': {
+    \       'callHierarchy': {
+    \           'dynamicRegistration': v:true,
+    \       },
```
hrsh7th, line 489: Sorry. The vim-lsp does not support dynamicRegistration yet. So it should be turned to `v:false`.

autoload/lsp.vim
```diff
@@ -548,12 +551,19 @@ function! lsp#default_get_supported_capabilities(server_info) abort
...
     \       'rangeFormatting': {
     \           'dynamicRegistration': v:false,
     \       },
     \       'references': {
     \           'dynamicRegistration': v:false,
     \       },
+    \       'rename': {
+    \           'dynamicRegistration': v:true,
+    \           'prepareSupport': v:true
+    \       },
```
hrsh7th, line 566: ditto for dynamicRegistration.

I think we also should add the following capabilities.

```
prepareSupportDefaultBehavior: 1
```

autoload/lsp.vim
```diff
@@ -576,7 +586,10 @@ function! lsp#default_get_supported_capabilities(server_info) abort
     \           'multilineTokenSupport': v:false,
     \           'serverCancelSupport': v:false
     \       },
-    \       'publishDiagnostics': {
+    \       'signatureHelp': {
+    \           'relatedInformation': v:true,
```
hrsh7th, line 590: ditto for dynamicRegistration.

hrsh7th:
LGTM.,I'll,merge,it.

PR #1412, opened 2023-01-04
Allow adjustable options of prop_add() for virtual text

ryuichiroh:
Hi maintainers ✋

I faced to a problem of changing the height and wrapping of lines frequency as shown in the video below.
And I found out the cause is that a `text_align` of `prop_add()` option is `"below"`.

*Before*

https://user-images.githubusercontent.com/121105839/210579403-00aefdd8-c0db-4162-812e-47d0a6f9429c.mov

So I improved that the `prop_add()` option can be specified via global variables.
For example:

```vim
let g:lsp_diagnostics_virtual_text_enabled = 1
let g:lsp_diagnostics_virtual_text_align = 'after'
let g:lsp_diagnostics_virtual_text_padding_left = 2
let g:lsp_diagnostics_virtual_text_wrap = 'truncate'
```

*After*

https://user-images.githubusercontent.com/121105839/210579502-4d3b6f2a-9c38-4c22-ab9f-f082c7581046.mov


mattn (commented):
LGTM

mattn:
Thank you

Issue #1415, opened 2023-01-12
Unexpected line break


ryicoh:
There is an unexpected line wrap between lines 5 and 6 as show in the image below.

<img width="831" alt="Screen Shot 2023-01-12 at 11 05 54" src="https://user-images.githubusercontent.com/37844673/211958617-fc58bb14-becd-4f56-957d-0d2621e77cf5.png">

## Reproducing

Minimum vimrc:

```vim
" vimrc_lsp.vim
set number
packadd vim-lsp
packadd vim-lsp-settings

let g:lsp_diagnostics_virtual_text_enabled = 1
let g:lsp_diagnostics_virtual_text_align = 'after'
let g:lsp_diagnostics_virtual_text_wrap = 'truncate'
autocmd BufNewFile,BufRead *.tsx set filetype=typescriptreact
```

Target code:
```bash
export function Document() {
  return "hello"
}

 have a good idea that add option to disable lsp_diagnostics_float_cursor in insert mode.
```

```bash
% vim -N -u vimrc_lsp.vim
```

Environment: 

```bash
% vim --version
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Dec 26 2022 13:51:26)
macOS version - arm64
Included patches: 1-1100
Compiled by Homebrew
```

ryicoh:
I will try to resolve it.

ryuichiroh:
I found that the cause is that both virtual text and diagnostic highlighting are enabled. So if disable highlighting, it will be able to be fixed.

```vim
" write to your vimrc
let g:lsp_diagnostics_highlights_enabled = 0
```

<img width="837" alt="Screen Shot 2023-01-13 at 0 36 38" src="https://user-images.githubusercontent.com/121105839/212110992-b6c39694-67b4-4b39-a020-f15ecb5c9d08.png">

ryuichiroh:
Line 5 has only 88 lines, but the `col` of `prop_add()` is 89!

```vim
" log
prop_add(5, 89, {'bufnr': 1, 'end_lnum': 5, 'type': 'vim_lsp_diagnostics_highlights_LspError', 'end_col': 89})
```

https://github.com/prabirshrestha/vim-lsp/blob/500987604d356738068ee3bf320a82dfa9fbfc1f/autoload/lsp/internal/diagnostics/highlights.vim#L167-L175

https://github.com/prabirshrestha/vim-lsp/pull/721

[...]

PR #1416, opened 2023-01-12
Check for valid range before place diagnostic highlights

ryu-ichiroh:
Close: #1415 

https://github.com/prabirshrestha/vim-lsp/pull/721#issuecomment-753279052

https://github.com/prabirshrestha/vim-lsp/blob/500987604d356738068ee3bf320a82dfa9fbfc1f/autoload/lsp/internal/diagnostics/highlights.vim#L167-L175

> f someone wants to create lsp#utils#range#is_valid_for_buffer(range, buffer) before calling prop_add feel free to send a PR.

I read this comment and I tried to add `lsp#utils#range#is_valid_for_buffer(range, buffer)`. But we have to check both start_line/start_col and end_line/end_col not range.

So I decided to add function `lsp#utils#position#is_valid_for_buffer(bufnr, line, col)`.

### DIFF
--- a/autoload/lsp/utils/position.vim
+++ b/autoload/lsp/utils/position.vim
@@ -89,3 +89,15 @@ function! lsp#utils#position#vim_to_lsp(expr, pos) abort
          \ }
 endfunction
 
+function! lsp#utils#position#is_valid_for_buffer(bufnr, line, col) abort
+    if a:col < 0
+        return 0
+    endif
+
+    let l:line_str = get(getbufline(a:bufnr, a:line, '$'), 0, '')
+    if a:col > strlen(l:line_str)
+        return 0
+    endif
+
+    return 1
+endfunction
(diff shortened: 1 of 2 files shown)


autoload/lsp/utils/position.vim
```diff
@@ -89,3 +89,15 @@ function! lsp#utils#position#vim_to_lsp(expr, pos) abort
          \ }
 endfunction
 
+function! lsp#utils#position#is_valid_for_buffer(bufnr, line, col) abort
+    if a:col < 0
+        return 0
+    endif
+
+    let l:line_str = get(getbufline(a:bufnr, a:line, '$'), 0, '')
+    if a:col > strlen(l:line_str)
```
mattn, line 98: better to use strdisplaywidth?

ryuichiroh: @mattn 

`col` parameter of `text_prop()` is counted in bytes. so I used `strlen()`.

> {col} is counted in bytes, use one for the first column.

https://github.com/vim/vim/blob/c0c2c262650103c4a21b64c3246388a350688616/runtime/doc/textprop.txt#L122-L126

Should I rename `lsp#utils#position#is_valid_for_buffer()`?

ryuichiroh:
I found a bug and will fix it.

PR #1420, opened 2023-01-21
Add option to show only virtual text of the highest severity

ryuichiroh:


hrsh7th:
I think it's useful but I also think it should be `let g:lsp_diagnostics_virtual_text_severities = [1,2,3...]`.

PR #1426, opened 2023-01-27
Make the CompleteDone backspace check more robust

infmagic2047:
When checking for backspace in `CompleteDone` event, also look at the line number. If the line number changed, the user probably pressed something like `<CR>` and not backspace.


hrsh7th:
Makes sense. Thank you.

PR #1468, opened 2023-04-23
fix eat of typing char after completion done

zbinlin:
When using typescript-language-server, in content `Math.ab|`, if you type `<C-X><C-o>` for omni completion, it will be completed with `s`. Then, the content will become `Math.abs|`. If you continue type `(`, the `(` will be removed. Also, if type `<CR>`, a new line will be added, resulting in `Math.abs|s\nMath.ab`. This patch attempts to fix this issue.


mattn:
Thanks. I'll look this soon

mattn:
Thank you

PR #1485, opened 2023-08-09
Fix disappeared kind

jiz4oh:
Hi @prabirshrestha,

the kind is disappeared now due to the incorrect ternary expression

image

refer #1482


jiz4oh:
it is expected after the pr image

Hi @prabirshrestha, could you have a look?

mattn (approved):

mattn:
Good catch. Thank you

jiz4oh:
Thanks @mattn

PR #1513, opened 2023-11-19
Fallback to the less efficient getbufline function when getbufoneline…

Jsbjr:
Fallback to the less efficient getbufline function when getbufoneline does not exists


Jsbjr:
Commit dbe8b17 fix it. Tis PR is not needed any more

PR #1537, opened 2024-02-21
Fix multiple diagnostic virtual text per-line of same severity in vim

dylanahsmith:
Remove the prop_remove just before prop_add in place_virtual_text, since this function is called after clear_virtual_text is called to clear the previously added virtual text.

That way, if there are multiple virtual text to place on the same line with the same severity, the following ones don't replace the former ones.  Besides, the branch for neovim wasn't removing any property, so this should make it consistent with neovim.


dylanahsmith:
I'm not sure what activity was expected, there are no merge conflicts to rebase to fix and I didn't receive any feedback on the pull request to address.

PR #1570, opened 2024-07-23
Diagnostic float: Record position before debounce time

ilya-bobyr:
It seems that we should be recording the current buffer, cursor position and change tick before we wait to see if the float should be shown or not.  Otherwise, if any of the above change during the `g:lsp_diagnostics_float_delay` our recording would be incorrect.


ilya-bobyr:
This is a first commit in [a sequence of 7 commits](https://github.com/prabirshrestha/vim-lsp/compare/master...ilya-bobyr:vim-lsp:master) that improve the float behavior for me.
Mostly by fixing some subtle bugs.

Each commit is documented, should work, and can be reviewed independently.
I think it makes sense to look at all of them eventually.
If you prefer, I can create a single PR for all of them combined.
Or I can create new PRs one by one.
Posting the next commit only after the previous commit goes through the review process.

All the changes are in the same area, so they rely on each other quite a lot.
To some degree, some changes are justified by what comes next.
But I was trying to split them such that the future dependencies are rather weak.

ilya-bobyr:
This is still an issue.
Would be great if someone could take a look.
I have 6 more changes on top of this one, that address a number of inconsistencies in with the diagnostic float behavior.

mattn:
LGTM

mattn:
Thank you

PR #1594, opened 2025-03-06
fix parsing header's Content-Length without carriage-return

nguyennk92:
Lsp client failed to read `Content-Length` from header. The server implementation was https://github.com/NomicFoundation/hardhat-vscode

Turns out the header response separator was `\n` instead of `\r\n`
image

Neovim lsp client doesn't have this issue, so I just apply this simple fix

### DIFF
--- a/autoload/lsp/client.vim
+++ b/autoload/lsp/client.vim
@@ -139,7 +139,7 @@ function! s:on_stdout(id, data, event) abort
 endfunction
 
 function! s:get_content_length(headers) abort
-    for l:header in split(a:headers, "\r\n")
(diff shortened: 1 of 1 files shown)


mattn:
FYI, according to the RFC, the header must be split by \r\n

nguyennk92:
You're right, seems to be bugged by the language server as they send the logs as part of response header. Don't know how neovim's lsp overcome this

I'll try to fix the language server then

nguyennk92:
My bad this was problem from our side.
Actually there's a `console.log` function called inside our `hardhat.config.ts`. 

image

This file seems to somehow getting executed from the language server

PR #1601, opened 2025-04-23
Add a setting for error message of LspHover

nami-minami:
Add a setting for error message of LspHover.
This setting can hide error message of LspHover that is "No hover information found in server - <server name>".


nami-minami:
I fixed commits and run a test on the themis. However, The test for lsp#uivim#completions was skipped and I did not know how to run it, so I did not run it. sorry...

PR #1604, opened 2025-05-24
Specify that Vim9 is needed for semantic highlighting

Sniliev:
Semantic highlighting uses reduce() and prop_type_add() which are introduced in Vim9.


mattn:
Hmm, sorry, I don't think this is required.

PR #1607, opened 2025-06-22
Handle create/rename/delete TextDocumentEdits

Lazzer64:
Not all [WorkspaceEdits](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit) are text document changes. Add support for create/rename/delete edits.


mattn:
Test seems still not working.
```
integration#go#text_document_edits

  [x] should create a new test file

      Error occurred line:

       16:         Assert Equals(expand("%:p:t"), "math_test.go")

      The equivalent values were expected, but it was not the case.

      

          expected: 'math_test.go'

               got: 'math.go'
```

PR #1610, opened 2025-08-10
Fix completion item text selection priority

heavenshell:
Hi,

I found vtslp and typescript-language-server complete invalid item.

Steps to reproduce
```typescript
type Address = {
  state?: string
  town?: string
}

type Person = {
  name?: string
  age?: number
  address?: Address
}

const foo: Person = {
  name: 'foo',
  age: 20,                 
}
```

When I type like following

```typescript
foo.address.<c-x><c-o>
            +--------------+
            | state? field |
            | town?  field |
            +--------------+
```
and select `state?`, text insert like following

```typescript
foo.address.?.state
```

expected behavior is

```typescript
foo.address?.state
```

According to the [Language Server Protocol 3.17 specification:](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion)

- filterText is only used for filtering and sorting completion items
- When textEdit is provided, insertText must be ignored

In vim-lsp completion context, l:completion_item['filterText'] was set.
I think this is unnecessary.

I would appreciate it if someone could review this.

Before

https://github.com/user-attachments/assets/21ce9284-78d9-4c99-aa88-6585f238d974

After

https://github.com/user-attachments/assets/33b02887-df46-4dcd-9d46-d7d51028ba0a

### DIFF
--- a/autoload/lsp/omni.vim
+++ b/autoload/lsp/omni.vim
@@ -312,15 +312,7 @@ function! lsp#omni#get_vim_completion_items(options) abort
         let l:range = lsp#utils#text_edit#get_range(get(l:completion_item, 'textEdit', {}))
         let l:complete_word = ''
         if has_key(l:completion_item, 'textEdit') && type(l:completion_item['textEdit']) == s:t_dict && !empty(l:range) && has_key(l:completion_item['textEdit'], 'newText')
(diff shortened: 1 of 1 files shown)


heavenshell:
This has been found that this has side effects.

PR #1627, opened 2025-12-31
Add end pos for qf and loclist

thinca:
The LSP response contains the end of the range, but it was not set in the quickfix or location list. This will be set.


mattn:
Thank you

PR #1636, opened 2026-01-12
Preserve completion doc scroll position

hbarcelos:
Fix completion documentation popup scrolling reset in Vim.

The completion docs popup is re-opened on `CompleteChanged`; this preserves the existing scroll position (topline) when the popup already exists, so manual scrolling doesn’t jump back to the top.

Tested on Vim 9.1 popup windows.

image

I also noticed that the `conceal` config was somewhat inconsistent, depending on how the user sets it. For reading purposes, it's better that `conceal` is always enabled, even when the line in the popover is highlighted.


mattn:
Thank you

PR #1666, opened 2026-04-05
Fix typo

zbinlin:


mattn:
Thanks but this should be generated via callbag.vim

```
:CallbagEmbed path=autoload/lsp/callbag.vim namespace=lsp#callbag
```

PR #1667, opened 2026-04-11
Skip calling complete() in omnifunc when there are no matches

bnnkw:
I want to use items from 'complete' sources when LSP has no completion items (for example, inside string literals or comments).
Before this change, the empty completion items from LSP closed the completion menu triggered by i_CTRL-N or 'autocomplete', and discarded those items.
 
After this change, the items from 'complete' sources stay visible when LSP returns no completion items.


mattn:
Thank you

PR #1629, opened 2026-01-01
Diagnostic float: Document event handing logic

illia-bobyr:
Current event handling seems somewhat confusing.  It is good to document the intent, in order to be able to asses if the code is doing what is intended.

PR #1669, opened 2026-04-18
Diagnostic float: Split event handling flow

### LINKED ISSUE #1510: Diagnostics hover and echo status message hidden by "--No lines in buffer--"
Steps to reproduce:

* Make sure a Python LSP is installed.
* `echo "foo()\nbar()" > /tmp/foo.py`
* `vim -u /tmp/minimal.vim /tmp/foo.py`
* When vim opens, move the cursor and see an error like `undefined name 'foo'` in a popup window and in the status message.  Both the window and the status disappear after a few seconds (maybe 5 seconds after moving the cursor?) and the status message becomes `--No lines in buffer--`.
* Expected behavior: diagnostics status message and the floating window stay visible until moving the cursor.

Curiously, if `g:lsp_diagnostics_float_delay` is `500` instead of `2000` or `1000`, the "No lines in buffer" message doesn't show up and the diagnostic status message stays in place, though the hover window still disappears after a few seconds.  (Is the floating window supposed to disappear?  I'd really like it to stay visible until I move the cursor.)  500 seems to be the magic value; anything larger results in "No lines in buffer" messages overwriting the status line while anything smaller doesn't.

Contents of `/tmp/minimal.vim`, assumes vim-plug is already installed:
```vim
set nocompatible

call plug#begin('~/.vim/plugged')
Plug 'prabirshrestha/vim-lsp'
Plug 'mattn/vim-lsp-settings'
call plug#end()

" Configuration for vim-lsp
let g:lsp_signs_enabled = 1
let g:lsp_diagnostics_echo_cursor = 1
let g:lsp_diagnostics_float_cursor = 1
let g:lsp_diagnostics_float_delay = 1000
let g:lsp_diagnostics_virtual_text_enabled = 0
let g:lsp_float_max_width = 0
```

This happens on the system vim for macOS 12.7 (9.0 with patches 1-1544), MacVim 9.0.1897, and a recent gvim on Debian Linux (9.0, don't have the patch list handy).  It's not limited to any particular LSP; I've seen it with the Python and Julia LSPs from `vim-lsp-settings` as well as a company-internal LSP at work.

### LINKED ISSUE #1620: Diagnostic floating window disappears automatically
When showing the diagnostic messages via the floating window, the floating window automatically disappears after ~5 seconds. This often does not given enough time to read the message. Currently there also isn't a way to configure the timeout or whehter the floating window should disappear.

illia-bobyr:
As different events actually need to trigger different actions, it is easier to follow the logic when each event has its own sequence of actions.

It also reduces the amount of work, as only |CursorMoved| can cause the float to appear.

Restore |CursorHold| functionality disabled by

```gitcommit
commit 10fac72582a7f95ec3a0bf6715aaf0a7eb997bec
Author: suguruwataru <23465321+suguruwataru@users.noreply.github.com>
Date:   Thu Jan 1 19:09:58 2026 +0800

    Stop `CursorHold` events from updating diagnostic floating window (#1624)

    Closes #1620

    Co-authored-by: mattn <mattn.jp@gmail.com>
```

It is still nice to have an option of hiding the float when cursor did not move long enough.  A subsequent commit adds a configuration option to control this behavior.

---

It is part of a sequence of changes to fix https://github.com/prabirshrestha/vim-lsp/issues/1510.
Previous change was https://github.com/prabirshrestha/vim-lsp/pull/1629.
This and subsequent commits in the sequence: https://github.com/prabirshrestha/vim-lsp/compare/474c656659b5fb51eec6770309e0211c8aa49b5b...illia-bobyr:vim-lsp:9372cfc2f69443d5c907e4e59124595f8689160c

### DIFF
--- a/autoload/lsp/internal/diagnostics/float.vim
+++ b/autoload/lsp/internal/diagnostics/float.vim
@@ -19,28 +19,10 @@ function! lsp#internal#diagnostics#float#_enable() abort
     " float and, potentially, show it later after a delay of
     " 'g:lsp_diagnostics_float_delay'.  So we always call 'hide_float()' for
     " this event, and may call 'show_float()' later.
-    "
-    " CursorHold - If the cursor did not move long enough in the normal mode, we
-    " want to hide the float.  So we want to call 'hide_float()'.  Because of
-    " the 'curpos' check, 'show_float()' is not going to be called, but we could
-    " be more explicit and just not call it at all.
-    "
-    " InsertEnter - If insert mode is entered and we do not show floats in the
-    " insert mode (g:lsp_diagnostics_float_insert_mode_enabled) we want to hide
-    " the cursor.  Similar to |CursorHold| we do not show the float due to the
-    " "mode() is# 'n'" check, but it seems more like an accident than the
-    " intended action.
-    let s:Dispose = lsp#callbag#pipe(
-        \ lsp#callbag#merge(
-        \   lsp#callbag#fromEvent(['CursorMoved']),
-        \   lsp#callbag#pipe(
-        \       lsp#callbag#fromEvent(['InsertEnter']),
-        \       lsp#callbag#filter({_->!g:lsp_diagnostics_float_insert_mode_enabled}),
-        \       lsp#callbag#tap({_->s:hide_float()}),
-        \   )
-        \ ),
-        \ lsp#callbag#filter({_->g:lsp_diagnostics_float_cursor}),
+    let s:Dispose_CursorMoved = lsp#callbag#pipe(
+        \ lsp#callbag#fromEvent(['CursorMoved']),
         \ lsp#callbag#tap({_->s:hide_float()}),
+        \ lsp#callbag#filter({_->g:lsp_diagnostics_float_cursor}),
         \ lsp#callbag#map({_->{
         \   'bufnr': bufnr('%'),
         \   'curpos': getcurpos()[0:2],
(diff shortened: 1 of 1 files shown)


mattn:
I don't make sure but it this can use merge & pipe like: https://github.com/prabirshrestha/vim-lsp/pull/1551#issuecomment-4016026455 ?

illia-bobyr:
> I don't make sure but it this can use merge & pipe like: [#1551 (comment)](https://github.com/prabirshrestha/vim-lsp/pull/1551#issuecomment-4016026455) ?

I think it makes is much harder to understand the logic when we merge events from 3 unrelated sources.
There seems to be no gain in doing that.
Yet it produces code that is pretty hard to follow for the reader.
You need to keep in mind all the possibilities and consider what happens in each of 3 cases at the same time.

This PR main goal is to remote the `merge` call.
There are actual bugs in this logic, that, I think, are very hard to see, exactly because three unrelated event flows were combined into one.

When we combine unrelated things, it also makes it harder to make changes.
The reason I'm fixing 1 bug in 7 commits is because I need to untangle these combined flows first.
And I'm trying to be very careful not to break existing logic.
Because it is so hard to follow, it is also hard to make sure that when I make a change to the insert mode behavior I would not break something in the normal mode behavior.

Moving to separate definitions of events for each case enables me to actually start fixing things with any level of confidence :)

illia-bobyr:
Maybe I missed your point.
Are you suggesting that it would allow us to dispose of the pipeline in one call if we combine them all with `merge` at the top level?
I would agree to that.
But I'm a bit worried that it would make it tempting to start merging things further, as was done in the previous version :)

I can do it if you insist.  But I'll be a little hesitant to do it myself.

mattn:
@prabirshrestha What do you think about this?

PR #1673, opened 2026-05-14
Add toggles for disabling auto open location/quickfix list

roumail:
Hi all,

First of all, Thank you for the great plugin. I use your plugin everyday and am very grateful. 

I am submitting this change set to get an idea from you if you think it would be a useful addition to your ecosystem. 

Just because you trigger a gd or gr, etc, shouldn't always move focus away from the window you're working on. I didn't want this and applied some hacks to prevent this from happening. I've now been using this for a couple of months and find it quite nice. 

The change needs the addition of a couple of global options to maintain compatibility with the current behaviour. 

This is actually my first ever PR on an open source project. Vimscript isn't even something that I would list as something I'm great at. Any feedback would be highly appreciate. You can also say this isn't a useful feature to add. That's fine. 

I haven't added any tests for this (I actually don't know how to do that - but hey, if this is something that's interesting to add to vim-lsp, I can dig further into the tests/ directory)

I'm just sharing this because I thought that maybe this is something useful for others too.

Rohail

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,16 @@ You can get detailed status on your servers using `:CheckHealth` with a plugin o
 if !has('nvim') | Plug 'rhysd/vim-healthcheck' | endif
 CheckHealth
 ```
+## Toggle location and quickfix behaviour
 
+The default behaviour is to open the location or quickfix list automatically
+when triggering a command.
+
+You may use the the configuration parameters below to prevent this from happening:
+```vim
+let g:lsp_auto_open_loclist = 0
+let g:lsp_auto_open_qflist = 0
+```
 ## Tests
 
 [vim-themis](https://github.com/thinca/vim-themis) is used for testing. To run
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ CheckHealth
 The default behaviour is to open the location or quickfix list automatically
 when triggering a command.
 
-You may use the the configuration parameters below to prevent this from happening:
+You may use the configuration parameters below to prevent this from happening:
 ```vim
 let g:lsp_auto_open_loclist = 0
 let g:lsp_auto_open_qflist = 0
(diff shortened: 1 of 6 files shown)


autoload/lsp/internal/diagnostics/document_diagnostics_command.vim
```diff
@@ -35,6 +35,6 @@ function! lsp#internal#diagnostics#document_diagnostics_command#do(options) abor
     else
         call setloclist(0, l:result)
         echo 'Retrieved diagnostics results'
-        botright lopen
+        call lsp#utils#maybe_execute_list_command('location', 'botright lopen')
```
roumail, line 38: I personally didn't like the "duplication in 'location and lopen'". You can derive the kind from the command directly.

However, this was just a first draft to get feedback

README.md
```diff
@@ -189,7 +189,16 @@ You can get detailed status on your servers using `:CheckHealth` with a plugin o
 if !has('nvim') | Plug 'rhysd/vim-healthcheck' | endif
 CheckHealth
 ```
+## Toggle location and quickfix behaviour
 
+The default behaviour is to open the location or quickfix list automatically
+when triggering a command.
+
+You may use the the configuration parameters below to prevent this from happening:
```
mattn: the the

roumail: Yeah can fix that!

mattn:
please add doc for `lsp_auto_open_loclist`.

autoload/lsp/utils.vim
```diff
@@ -195,6 +195,20 @@ function! lsp#utils#get_buffer_uri(...) abort
...
+    endif
+  elseif a:kind ==# 'quickfix'
+    if !get(g:, 'lsp_auto_open_qflist', 1)
+      return
+    endif
+  endif
+
+  execute a:cmd
+endfunction
+
```
mattn, line 211: please fix indent with 4 spaces.

roumail:
> please add doc for `lsp_auto_open_loclist`.

This is the section in the readme where I document both the parameters together
```
The default behaviour is to open the location or quickfix list automatically
when triggering a command.

You may use the configuration parameters below to prevent this from happening:
```vim
let g:lsp_auto_open_loclist = 0
let g:lsp_auto_open_qflist = 0
```
I'm sorry, but is there something specific you think is missing from the description above? Happy to adapt!

PR #1675, opened 2026-05-21
Improving multiple workspaces functionality

nonrice:
Appears the wrong schema is used in `send_notification` for when the workspace folder is changed? Fixed it.

Also added public methods for explicitly managing workspace folders. As a reason for inclusion: For me they were necessary for managing some tooling at my job. These methods were mentioned in #1265 but not implemented yet, maybe due to the lack of necessity?

I tested these changes with some basic work on java (jdtls) and rust (bacon-ls/rust-analyzer) projects with [mattn/vim-lsp-settings](https://github.com/mattn/vim-lsp-settings), both with multiple workspaces. Existing functionalities seem fine.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -722,7 +722,7 @@ function! s:ensure_init(buf, server_name, cb) abort
         let l:root_uri = lsp#utils#get_default_root_uri()
     endif
     let l:server['server_info']['_root_uri_resolved'] = l:root_uri
-    let l:server['workspace_folders'][l:root_uri] = { 'name': l:root_uri, 'uri': l:root_uri }
+    let l:server['workspace_folders'][l:root_uri] = { 'name': lsp#utils#uri_to_path(l:root_uri), 'uri': l:root_uri }
 
     if has_key(l:server_info, 'capabilities')
         let l:capabilities = l:server_info['capabilities']
@@ -744,9 +744,8 @@ function! s:ensure_init(buf, server_name, cb) abort
     
     let l:workspace_capabilities = get(l:capabilities, 'workspace', {})
     if get(l:workspace_capabilities, 'workspaceFolders', v:false)
-        " TODO: extract folder name for l:root_uri
         let l:server_info['workspaceFolders'] = [
-            \ { 'uri': l:root_uri, 'name': l:root_uri }
+            \ { 'uri': l:root_uri, 'name': lsp#utils#uri_to_path(l:root_uri) }
             \ ]
         let l:request['params']['workspaceFolders'] = l:server_info['workspaceFolders']
     endif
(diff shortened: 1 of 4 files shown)


mattn:
The existing `s:workspace_add_folder()` (at `autoload/lsp.vim:909`) bails out when `g:lsp_experimental_workspace_folders` is off, and `initialize` (line 692) only advertises `workspaceFolders: true` when that flag is set. The new `lsp#add_workspace_folder` / `lsp#remove_workspace_folder` skip this check, so if a user calls them while the flag is off, we send `didChangeWorkspaceFolders` to a server that was told the client doesn't support workspace folders. Could you add the same guard (or an `lsp#utils#error`) to the new functions?

autoload/lsp.vim
```diff
@@ -1415,6 +1418,104 @@ function! lsp#update_workspace_config(server_name, workspace_config) abort
...
+    if !lsp#is_server_running(a:server_name)
+        call lsp#utils#error('Server is not running: ' . a:server_name)
+        return
+    endif
+    if !lsp#capabilities#has_workspace_folders_change_notifications(a:server_name)
+        call lsp#utils#error('Server does not support workspace folder notifications: ' . a:server_name)
+        return
+    endif
+
+    let l:uri = a:path_or_uri =~# '^file://' ? a:path_or_uri : lsp#utils#path_to_uri(a:path_or_uri)
```
mattn: `a:path_or_uri =~# '^file://'` only matches `file://`. Switching to `=~# '^\w\+://'` would be safer for future schemes. Not blocking.

autoload/lsp.vim
```diff
@@ -1415,6 +1418,104 @@ function! lsp#update_workspace_config(server_name, workspace_config) abort
...
+    let l:server = s:servers[a:server_name]
+
+    if has_key(l:server['workspace_folders'], l:uri)
+        return
+    endif
+
+    let l:workspace_folder = { 'name': lsp#utils#uri_to_path(l:uri), 'uri': l:uri }
```
mattn, line 1445: - Existing: `{ 'name': root_uri, 'uri': root_uri }`
   - New: `{ 'name': uri_to_path(uri), 'uri': uri }`

   The new form is closer to the LSP spec (human-readable name), but mixing the two within the same server's folder list is confusing. Please pick one (I'd prefer the path-based form, and update the existing call site too).

autoload/lsp.vim
```diff
@@ -1415,6 +1418,104 @@ function! lsp#update_workspace_config(server_name, workspace_config) abort
...
+        \    },
+        \  }
+        \ })
+    unlet l:server['workspace_folders'][l:uri]
+
+    let l:server_info = l:server['server_info']
+    if has_key(l:server_info, 'workspaceFolders')
+        call filter(l:server_info['workspaceFolders'], {idx, val -> val['uri'] !=# l:uri})
+    endif
+endfunction
```
mattn, line 1507: `add_workspace_folder` silently returns if the folder is already registered, while `remove_workspace_folder` errors when missing. Either align the behavior or add a comment explaining the intent.

mattn:
- Please add `LspAddWorkspaceFolder` / `LspRemoveWorkspaceFolder` and the new autoload functions to `doc/vim-lsp.txt`.
- A test that asserts the new `event`-wrapped notification payload would help guard against regressions of the schema fix.

nonrice:
Not sure the best way to add a test for the schema, could I get help with this? I believe the other comments are appropriately addressed now, though.

PR #1682, opened 2026-07-22
Improve interoperability under win32unix

ktahar:
## Problem

  Under `win32unix` (Cygwin/MSYS), vim and a language server may use
  different path and PID namespaces. This caused problems with native
  Windows language servers such as [basedpyright](https://github.com/detachhead/basedpyright):

  - vim-lsp generates Windows-style file URIs under `win32unix`, but `normalize_uri()` did not normalize them. 
    URI differences such as drive-letter casing or an encoded drive colon could therefore prevent
    responses from matching the corresponding vim buffer.
  - `getpid()` returns a Cygwin/MSYS PID that may not identify vim to a native Windows server. 
    A server using it for parent-process monitoring may incorrectly conclude that vim has exited.
  - `rootPath` uses a platform-dependent filesystem path and may contradict the Windows-style `rootUri`.

  ## Changes

  - Apply `normalize_uri()` under `win32unix`.
  - Only lowercase a drive letter when it is followed by a colon, avoiding
    changes to paths such as `file:///Home/project`.
  - Send `processId: null` under `win32unix`, where the server's PID
    namespace cannot be determined reliably.
  - Omit the deprecated `rootPath` under `win32unix` and rely on
    `rootUri`. It remains present on other platforms for compatibility.
  - Run URI normalization tests under `win32unix` and skip native Windows
    and Unix path-conversion fixtures that do not apply there.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -733,15 +733,17 @@ function! s:ensure_init(buf, server_name, cb) abort
     let l:request = {
     \   'method': 'initialize',
     \   'params': {
-    \     'processId': getpid(),
+    \     'processId': has('win32unix') ? v:null : getpid(),
     \     'clientInfo': { 'name': 'vim-lsp' },
     \     'capabilities': l:capabilities,
     \     'rootUri': l:root_uri,
-    \     'rootPath': lsp#utils#uri_to_path(l:root_uri),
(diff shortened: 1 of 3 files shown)


ktahar:
Please let me withdraw this patch. It could be premature.

Issue #1684, opened 2026-08-03
tsc --lsp (TypeScript v7) does not work with vim-lsp


heavenshell:
Hi,

TypeScript v7 ships `tsc --lsp --stdio`, which is a native LSP server built from [microsoft/typescript-go](https://github.com/microsoft/typescript-go). 
It starts faster and checks faster than `typescript-language-server` or `vtsls`, but some vim-lsp issues prevent it from working.

## `client/registerCapability` is not supported

tsc sends `client/registerCapability` right after `initialize`, but vim-lsp does not respond to it,
so tsc blocks and every subsequent message is dropped.

## native client can't send responses with a string id

When the server sends a request with a string id such as `"id": "ts1"` (tsc uses these for `workspace/configuration` and similar),
the client fails to send back a response with that id under `g:lsp_use_native_client = 1`. 
`ch_sendexpr` raises `E475: Invalid value for argument id` and the response never reaches  the server.

This behavior is Vim's current LSP channel design, only numeric ids are supported(see `:h channel-lsp` states "Vim currently supports only a number type for the id field"),
and this limitation is tracked in [vim/vim#14091](https://github.com/vim/vim/issues/14091).
The [LSP spec (RequestMessage)](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) allow the id to be `integer | string`.                                                                                                                                          

## Reproduction

Minimal `vimrc`:

```vim
set nocompatible
set runtimepath^=PATH/TO/vim-lsp
runtime plugin/lsp.vim
filetype plugin indent on

augroup MyLspSetup
  autocmd!
  autocmd User lsp_setup call lsp#register_server({
    \ 'name': 'tsc',
    \ 'cmd': {server_info -> ['/PATH/TO/node_modules/.bin/tsc', '--lsp', '--stdio']},
    \ 'allowlist': ['typescript', 'typescriptreact'],
    \ })
augroup END
```

I'll create two pull requests to solve these issues.

Thank you.

KiYugadgeter:
Any progress?

heavenshell:
I already submitted a pull request.
Have you tried applying the patch and reporting whether it works or not?

mattn:
Merged pull-requests.

heavenshell:
Thank you for merging

PR #1685, opened 2026-08-03
Fix respond to client/registerCapability and client/unregisterCapability

heavenshell:
ref: https://github.com/prabirshrestha/vim-lsp/issues/1684

Support `client/registerCapability` / `client/unregisterCapability` because TypeScript v7's `tsc --lsp` requires it.


mattn:
Thank you

